PR #10, opened 2018-02-17
Example project to show how mutation testing should be used

hmvp:
This is a preliminary version of a pull request to add some example code. 

It is not finished yet but I don't really see how to proceed. But I want to post it here to get some feedback. Some of these questions should be answered in the readme or a how to use guide.

## Open questions

- The mutate attribute: should I place it above all functions that I want to mutate during test?
- Can/should I make that conditional on cfg(test)? Not sure if I want mutagen code in the final binary
- I made the first test always fail to show the println!s It seems to show that mutagen is doing something. Is this the case?
- I used MU.next() 100 times, is there a max or does it depend on the number or types of mutations that are applied? It does not seem to change much after 10 times..
- Should there be a `#[mutate_test]` attribute that you could use instead of `#[test]` to create tests with mutagen?

## TODO

- [x] The example code is copied from a project without license. I should write a different example to clear up the license issue. I was lazy
- [x] The first test tries to show that mutagen is working. I am not sure though if this is the intended purpose or direction. But the intention is to make a start with a manual mutagen runner (and maybe move that into the lib/plugin?


llogiq:
Thanks for the PR! I plan to write out the list of mutations; the maximum mutation count should be the length of this list + 1. The `#[cfg(test)]` is a good idea, we should absolutely suggest that in the README. I don't think we should require a `#[mutagen_test]` attribute – at the end of the day they're just tests.

Apart from the license issue, I'd be OK with merging this as is. We'll figure out this stuff as we go along.

otavio:
> I don't think we should require a #[mutagen_test] attribute – at the end of the day they're just tests.

Indeed but avoiding boilerplate is always good and a mark also helps to identify which ones are using the feature.

Possibly, it could use:

```rust
#[test]
#[mutagen]
fn my_test() {
    // code
}
```

examples/example_project/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+    use super::mutagen;
+
+    #[test]
+    fn fizz_for_multiples_of_3() {
+        for _ in 0..100 {
+            mutagen::MU.next();
+
+            println!(
+                "Test {}: {} == {} => {}",
+                mutagen::MU.get(),
```
otavio: The example should use the new methods here, no?

llogiq: Yes, it should.

llogiq:
I'd rather name the tests on the command line than fixing them in code.

Keirua:
Great! I also felt the need to have a sample project in order to understand how the project runs. I just started a similar thing, but I'm happy to see it's already been done!

llogiq:
So are there any further license issues with the example that you know of?

One thing that I'm planning is to change the mutation count to something that can be set with an environment variable, so we can set it from outside the code, however this would render your example uncompilable. I *think* it should be possible to support both styles (which would be good for ...testing), but I'm not yet sure how to do it.

hmvp:
Not that I know. I removed all the copied code so everything in here is either generated by cargo or written by me

llogiq:
:+1:

llogiq:
Thanks again!

Issue #5, opened 2018-02-16
Mutation: Change literals


llogiq:
A simple, but effective mutation is to introduce off-by-one errors by changing numeric literals. Obviously we want to care about the type so we don't end up with e.g. `-1usize` or `256u8`, but we should be able to check this.

gnieto:
Closed via https://github.com/llogiq/mutagen/pull/26

PR #26, opened 2018-02-23
Add off-by-one mutation

gnieto:
As suggested on https://github.com/llogiq/mutagen/issues/5, it would be
nice to mutate numeric constants adding or subtracting one unit to numeric constants.

At the moment, this is very conservative on the limits to add or
substract on default integer constants. I've did it like this to avoid
compilation issue, as we do not have information regarding the type of
the expression. So, in the default case, it subtracts one only if it's >
0 and it only adds one if the value is < 256.

If the constant has a type annotation (for example, 0i32), It adjust the
minimum and the maximum depending on the type hint. Note that it does
not check 128 bits numbers, as they are only available on nightly.

It chains the mutations in order to record only the needed
mutations. So, if the code finds a "0" on an unsigned type, it will only
record the "add one" mutation, but not the "sub one" (as it would overflow and panic).

After running it on the example project, it emits mutations like:

```
    let mut count = {
        if mutagen::now(2usize) {
            0 + 1
        } else {
            0
        }
    };
```
or

```
              ord < {
                    if mutagen::now(12usize) {
                        65 + 1
                    } else {
                        {
                            if mutagen::now(11usize) {
                                65 - 1
                            } else {
                                65
                            }
                        }
                    }
                }
```

And finally: I didn't add tests. I thought on adding another test on the example project that covers some of the edge cases and I've also thought on unit test the `int_constant_can_subtract_one` or `int_constant_can_add_one`, but I'm not sure which is the expected way of testing this library, so, if you have some idea, I can add the kind of tests you think will be more useful.

### DIFF
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -16,7 +16,7 @@ use syntax::fold::{self, Folder};
 use syntax::ptr::P;
 use syntax::symbol::Symbol;
 use syntax::util::small_vector::SmallVector;
-use syntax::ast::{LitKind, LitIntType, IntTy};
+use syntax::ast::{LitKind, LitIntType, IntTy, UnOp};
 
 #[plugin_registrar]
 pub fn plugin_registrar(reg: &mut Registry) {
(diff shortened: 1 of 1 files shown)


plugin/src/lib.rs
```diff
@@ -473,12 +474,121 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
...
+        _ => std::i64::MIN,
+    };
+
+    i as i64 > min
+}
+
+fn int_constant_can_add_one(i: u64, ty: LitIntType) -> bool {
+    let max: u64 = match ty {
+        LitIntType::Unsuffixed => 256,
+        LitIntType::Unsigned(UintTy::Usize) => std::u32::MAX as u64,
```
gnieto: I doubt which should be the max value for usize. As far as I know, it's system dependent, so I'm not sure which is the most safe value here.

plugin/src/lib.rs
```diff
@@ -473,12 +474,121 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
...
+        LitIntType::Signed(IntTy::I64) => std::i64::MIN as i64,
+        _ => std::i64::MIN,
+    };
+
+    i as i64 > min
+}
+
+fn int_constant_can_add_one(i: u64, ty: LitIntType) -> bool {
+    let max: u64 = match ty {
+        LitIntType::Unsuffixed => 256,
```
gnieto: And I also have doubts here. Maybe I should change it to std::i8::MAX, as it's the max positive value that won't cause compilation issues?

plugin/src/lib.rs
llogiq: This shouldn't be a problem as long as we're consistent.

plugin/src/lib.rs
llogiq: Agreed, let's be careful here.

llogiq:
Great work so far! I was just thinking about implementing this mutation and you beat me to it :smile:

Apart from small nits, this looks good to merge to me. If you like, feel free to add unit tests, but I won't block on them now.

gnieto:
I've been checking this again and I've found that it does not properly work with negative values.
For example with the following code:
```
let test: i32 = -24322;
```

I get: `LitKind(24322, LitIntType::Unsuffixed)` . I'll look how I can workaround this.

llogiq:
This is because you have an UnOp(UnNeg, Lit(..)) – working around will require folding UnOps and looking into whether the contained `Expr` is a `Lit`, then mutating with a 'negative' flag (which means it'll be an `i*` type).

llogiq:
@gnieto Do you want to extend this or should I merge as is and do it myself? I'll be happy in either case.

gnieto:
If you don't mind, i would like to finish it. I expect to finish it late
today or tomorrow with the negative literal handling. Thanks!

El dia 25 febr. 2018 15:10, "llogiq" <notifications@github.com> va escriure:

@gnieto <https://github.com/gnieto> Do you want to extend this or should I
merge as is and do it myself? I'll be happy in either case.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/llogiq/mutagen/pull/26#issuecomment-368312218>, or mute
the thread

llogiq:
Great! No offense meant. I just wanted to make sure the PR doesn't become stale. :+1:

llogiq:
Great work! Thanks again! :+1:

PR #30, opened 2018-02-26
Add testing framework

gnieto:
Adding a minimal testing framework, which will allow to test the plugin
and the mutations generations.

The rationale behind is create integration tests with one or more blocks
with the #[mutate] annotation. As the test will be compiled before
executing, when the tests will be excuted, we will be able to find the
`mutations.txt` file generated for this specific integration test. The idea is that we may have one integration test per logical set of mutations. On this example, I'm adding an example for binary operations.

Then, we can assert one or more messages with should be located on some
specific line. Note that we should be able to check also that some
mutation is not generated (for example: `assert_eq!(false,
checker.has("replacing _ || _ with false", "12"))`, will check that
this mutation is not being applied on line 12).

The tests will be fragile, as they will fail if:
1- Mutation output format is changed
2- If the spans of the testing example is changed
3- If any message is changed


gnieto:
I don't know why the build has not failed, as the log shows clearly an error. We will need to fix the `.travis.yml` file in order to make it fail.

.travis.yml
```diff
@@ -5,8 +5,8 @@ os:
   - osx
 cache: cargo
 sudo: false
-script: |
-  cargo test
-  pushd plugin
-  cargo test
-  popd
+script:
```
gnieto: Without this change, it seems that only the status code of the last command (popd) is used. If the 2nd `cargo test` failed, the build was passing, as it's status code was shadowed by the `popd` one.

llogiq: Interesting. Good catch :+1:!

llogiq:
Thank you! This is great to have.

PR #31, opened 2018-02-26
Fix eq and ne mutations 

gnieto:
The code generated for eq and non-eq binary operations was not correct.
There was an issue with the generated code.

Previously, the expression generated on the macro, where wrapping the
expression on a closure, but the closure does not implement PartialEq,
which is needed to be able to compare for equality/non-equality.

The code has been changed to receive a boxed closure which returns a
PartialEq. This is the unique way I've found to make an easy example
compile.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -124,22 +124,24 @@ impl Mutagen {
     }
 
     /// use instead of `==`
-    pub fn eq<T: PartialEq>(&self, x: T, y: T, n: usize) -> bool {
+    pub fn eq<X, Y, T: PartialEq>(&self, x: X, y: Y, n: usize) -> bool
+    where X: FnOnce() -> T, Y: FnOnce() -> T {
         match n.wrapping_sub(self.get()) {
             0 => true,
             1 => false,
-            2 => x != y,
-            _ => x == y
+            2 => x() != y(),
+            _ => x() == y()
         }
     }
 
     /// use instead of `!=`
-    pub fn ne<T: PartialEq>(&self, x: T, y: T, n: usize) -> bool {
+    pub fn ne<X, Y, T: PartialEq>(&self, x: X, y: Y, n: usize) -> bool
+        where X: FnOnce() -> T, Y: FnOnce() -> T {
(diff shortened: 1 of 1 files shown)


llogiq:
I just came up with an even better idea: Use boolean operators to get rid of the closures altogether: `mutagen::now($n) || mutagen::now($n +1) && mutagen::now($n + 2) != ($left == $right)`

gnieto:
The solution was much more easy, I guess. I've changed the fix by just removing the closures arrround eq and ne: 65bed02665871a23fd658e082faf51908c02b859

The error I was getting was this one:
```
error[E0277]: the trait bound `[closure@tests/binops.rs:16:32: 16:36]: std::cmp::PartialEq` is not satisfied
  --> tests/binops.rs:16:20
   |
16 |     if mutagen::t((mutagen::eq(|| 2, || 3, 18usize)), 15usize) {}
   |                    ^^^^^^^^^^^ can't compare `[closure@tests/binops.rs:16:32: 16:36]` with `[closure@tests/binops.rs:16:32: 16:36]`
   |
   = help: the trait `std::cmp::PartialEq` is not implemented for `[closure@tests/binops.rs:16:32: 16:36]`
   = note: required by `mutagen::eq`
```

There was any reason on having that closures? I'm missing something?

llogiq:
There's a subtle distinction: Evaluation. With both the closures and the `&&`/`||` shenanigans the left and right side are not evaluated unless needed. This may actually make a difference if they have side-effects, e.g. `foo(stack.pop()) && bar(stack.pop())`. Evaluating the functions no matter what would change semantics in those cases, and we should be careful not to do that.

I wonder if we should split the mutation into "don't evaluate at all" and "evaluate and ignore the result".

gnieto:
Ok, I think i finally got it.

I've maintained the closure, I've fixed the type of the closure and I'm invoking the closure when it's needed (as the "and" or "or" mutation).
I've also added a test to ensure that it works on expressions which has side-effects.

Hopefully, this will solve all the issues.

llogiq:
Looks good to me.

Issue #2, opened 2018-02-13
programmatically run the test suite


llogiq:
The easiest way is to run `cargo test` with the mutagen plugin, then run the resulting binary for each mutation, giving the mutation number to the test binary as an environment variable.

* [x] compile tests with plugin
* [x] read mutation list, call test binary with mutation count

Keirua:
From what I understand, in a first step that sounds like a shell script, right ? It seems to depend on https://github.com/llogiq/mutagen/issues/15

llogiq:
Yes. That and finding out (or setting) the correct name for the test binary.

llogiq:
I asked /r/rust, `cargo test --no-run --message-format=json` will output the name of the executable, among other things.

PR #35, opened 2018-03-03
WIP Add mutagen test runner

hmvp:
Not finished yet but wanted the put it out here early to get feedback. This should solve #2.

This runs cargo test for every mutation and checks if at least one test fails

The idea is to run `cargo install` in the runner dir then run `cargo test && cargo mutagen` in the example dir

TODO:

- [ ] Cleanup unwraps
- [ ] Make sure the tests are compiled from the runner (you currently need to do that beforehand)
- [ ] Kill long running test suites
- [ ] Update example to only fail on one mutation as an example. It currently failes on many

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 # will have compiled files and executables
 /target/
 /plugin/target/
+/runner/target/
 
 # Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
 # More information here http://doc.crates.io/guide.html#cargotoml-vs-cargolock
--- a/README.md
+++ b/README.md
@@ -22,8 +22,6 @@ This project is basically an experiment to see what mutations we can still apply
 
 ### Running mutagen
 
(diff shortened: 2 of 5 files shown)


hmvp:
Hmm something does not work as expected. The return default mutation passes all test while doing that manually does not pass. But other mutations do fail one or more tests...

llogiq:
For the record, I don't worry at all about the unwraps. I had already started w/ code to build and find the test binary from the runner, so I think we'll best just merge this and sort the other things out later.

llogiq:
So, guess what I'm saying is: I don't want to merge a PR marked with [WIP], but feel free to merge whenever you're ready to – no need for it to be perfect now.

hmvp:
It turns out its already quite ok. Will save the rest of the todos for later. Especially to give you a chance to merge your changes.

There are a few mutations that don't seem to trigger failing tests even though they should. I will file some bugs (or fix them) if I can find the root cause..

PR #38, opened 2018-03-04
Use root namespace on macro expansion

gnieto:
When mutating code located on a module (so, any code which is not the one which has `extern crate mutagen;`), it fails to compile the code generated from the plug-in. It throws the following error:

```
    |
164 |     #[cfg_attr(test, mutate)]
    |                      ^^^^^^^ Use of undeclared type or module `mutagen`
```

To fix this issue, I have changed the macro expansions to reference the
mutagen functions from the root namespace.


llogiq:
Thank you!

PR #39, opened 2018-03-05
Change the runner to find and run the test binary directly

llogiq:
This parses the message output of `cargo test --no-run --message-format=json` to find out the location of the test binary to call it directly without the indirection through `cargo test`, which should be a bit faster.

I also added myself to the author list so I can publish cargo-mutagen on crates.io.

### DIFF
--- a/runner/Cargo.toml
+++ b/runner/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "cargo-mutagen"
 version = "0.1.0"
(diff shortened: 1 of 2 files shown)


hmvp:
Wow this makes it a lot faster..  0.5s vs 10s..

llogiq:
Yeah, running `cargo test` for each mutation is quite wasteful, because it has to check all sources (and possibly dependencies).

Issue #32, opened 2018-02-26
Apply mutations recursively


gnieto:
Currently the plugin is not applying rules recursively. For example, if the following snippet needs to be mutated:

```
if count == 2 {
    // Some
}
```

Generates the following code:

```
    if mutagen::t(
        count == {
            if mutagen::now(29usize) {
                2 + 1
            } else {
                {
                    if mutagen::now(28usize) {
                        2 - 1
                    } else {
                        2
                    }
                }
            }
        },
        25usize,
    ) {
        // Some!
    }
```

But if the snippet is:

```
    if (count == 2) {
        // Some!
    }
```

The generated code will be:

```
    if mutagen::t((mutagen::eq(|| count, || 2, 28usize)), 25usize) {
        // Some!
    }
```

On the first snippet, the "if" and the "off-by-one" mutations are applied.
On the second one, the "if" and the "eq" are applied.

Note that the plugin should be able to apply the 3 mutations on any of the cases. This should be as easy as change the calls to `fold::noop_fold_expr` to `self.fold_expr` instead. This should be changed on the `fold_expr`.

llogiq:
Ah, I see `noop_fold_expr(expr, self)` calls `self.fold_expr(_)` on all *sub*expressions of `expr`. So we should indeed issue `self.fold_expr(_)` on all sub-expressions of whatever we fold directly.

PR #40, opened 2018-03-05
Apply mutations recursively.

manaskarekar:
WIP Address issue #32.

https://github.com/llogiq/mutagen/issues/32

### DIFF
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -596,7 +596,7 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
 
             },
             e => {
-                P(self.fold_expr(P(e)))
+                P(fold::noop_fold_expr(e, self))
             },
         }) //TODO: more expr mutations
     }


manaskarekar:
Hi, I was having trouble getting the tests working before making and changes so I have been unable to test this.

plugin/src/lib.rs
```diff
@@ -596,7 +596,7 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
 
             },
             e => {
-                P(fold::noop_fold_expr(e, self))
+                P(self.fold_expr(P(e)))
```
llogiq: This one should stay as it is, lest we get an infinite loop.

manaskarekar: Ah, of course!

manaskarekar:
I would prefer to add the tests. 

Could you help me out on what I'm doing wrong, I can't seem to get the tests running (even before making my changes).

I tried `cargo +nightly test` and `cargo test` on the root and plugin folder but both failed.

Thanks!

gnieto:
@manaskarekar If I'm not wrong, on master the build is fixed, so if you rebase from master (or merge) tests should be ok.

Then, if you do what I suggested, take into account that you will add the test on the "plugin" project, so, you should run `cargo test` on the "plugin" folder.

manaskarekar:
Thanks, I have to step away for a bit. I'll add them shortly!

manaskarekar:
I add the test as you mentioned but I'm seeing the following as in debug

> "replacing if condition with false @ tests/binops.rs:30:8: 30:16",
"replacing if condition with true @ tests/binops.rs:30:8: 30:16",
"inverting if condition @ tests/binops.rs:30:8: 30:16",
"sub one to int constant @ tests/binops.rs:30:15: 30:16",
"add one to int constant @ tests/binops.rs:30:15: 30:16"

Is that the expected output? Doesn't show me the "replacing _ == _ with false".

I'm going to commit what I have so far, sorry haven't had a chance to dive in properly.

manaskarekar:
Here's a quick diff of what I added:

`--- a/plugin/tests/binops.rs
+++ b/plugin/tests/binops.rs
@@ -26,6 +26,11 @@ fn mutated_function() {
     if (2 != 3) {
         // Do something
     }
+
+    if ord == 2 {
+        // Some
+    }
+
 }
 
 #[test]
@@ -83,4 +88,34 @@ fn test_binop_ne() {
     ];
 
     assert!(checker.has_multiple(noneq_msgs, "26:9: 26:15"));
+}
+
+#[test]
+fn test_binop_eq_and_off_by_one() {
+    let checker = MutationsChecker::new("tests/binops.rs").unwrap();
+
+    //let eq_msgs1 = &[
+    //    "sub one to int constant",
+    //    "add one to int constant",
+    //    "inverting if condition",
+    //    "replacing _ == _ with false",
+    //    "replacing _ == _ with true",
+    //];
+
+    let eq_msgs = &[
+        "inverting if condition",
+        "replacing if condition with false",
+        "replacing if condition with true",
+    ];
+
+    assert!(checker.has_multiple(eq_msgs, "30:8: 30:16"));
+
+
+    let eq_msgs = &[
+        "sub one to int constant",
+        "add one to int constant",
+    ];
+
+    assert!(checker.has_multiple(eq_msgs, "30:15: 30:16"));
+
 }
`

plugin/src/lib.rs
```diff
@@ -319,8 +319,8 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
                             ],
                         );
                     }
-                    let left = left.map(|e| fold::noop_fold_expr(e, self));
-                    let right = right.map(|e| fold::noop_fold_expr(e, self));
+                    let left = left.map(|e| self.fold_expr(P(e)));
```
gnieto: if you run `cargo +nightly test` on the "plugin" folder you will see that this line is not compiling. If I'm not wrong, you need to change it to:

```rust
let left = self.fold_expr(left);
```
and the same with the other changes.

Note that `fold::noop_fold_expr` expects `Expr`, and because of this, the `map` was needed. On the other hand, `self.fold_expr` expects a `P(Expr)` and returns `P(Expr)`, so the map is not needed.

plugin/src/lib.rs
```diff
@@ -512,9 +512,9 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
                         ]
                     );
                 }
-                let cond = cond.map(|e| fold::noop_fold_expr(e, self));
+                let cond = cond.map(|e| self.fold_expr(P(e)));
                 let then = fold::noop_fold_block(then, self);
-                let opt_else = opt_else.map(|p_else| p_else.map(|e| fold::noop_fold_expr(e, self)));
+                let opt_else = opt_else.map(|p_else| p_else.map(|e| self.fold_expr(P(e))));
```
gnieto: Here you have a double map. If I'm not wrong, this should be change to:

```rust
let opt_else = opt_else.map(|p_else| self.fold_expr(p_else));
```

manaskarekar:
Sorry, I deleted my fork!

I'll just start over. I made the relevant changes and they are failing where you said it would due to the other issue #43.

llogiq:
Ok, since you want to start over, I guess we can close this, right?

PR #43, opened 2018-03-05
Fix unequality operands

gnieto:
Related to: https://github.com/llogiq/mutagen/pull/31
Blocks: https://github.com/llogiq/mutagen/pull/40

### DIFF
--- a/plugin/Cargo.toml
+++ b/plugin/Cargo.toml
@@ -13,7 +13,7 @@ crate-type = [ "dylib" ]
 plugin = true
 
 [dependencies]
-mutagen = "0.1.0"
+mutagen = {"version" = "0.1.0", "path" = ".."}
 
 [badges]
 travis-ci = { repository = "llogiq/mutagen", branch = "master" }
--- a/plugin/tests/binops.rs
+++ b/plugin/tests/binops.rs
@@ -26,6 +26,10 @@ fn mutated_function() {
     if (2 != 3) {
         // Do something
     }
+
+    if (ord < 2) {
+        // Do something
+    }
 }
 
 #[test]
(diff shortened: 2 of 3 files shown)


plugin/Cargo.toml
```diff
@@ -13,7 +13,7 @@ crate-type = [ "dylib" ]
 plugin = true
 
 [dependencies]
-mutagen = "0.1.0"
+mutagen = {"version" = "0.1.0", "path" = ".."}
```
gnieto: I've doubts here: @llogiq you change this line few days ago. How this should work? Should we avoid this and wait to publish another version of the lib? It would make sense to merge lib and plugin library?

llogiq: I think we should bump the version and publish all crates.

I'd also suggest to keep the version numbers of all crates in sync, to keep things simple.

gnieto:
@llogiq Care to elaborate? How this case is distinct from the `eq` or `ne` case? I'm doing this change because, currently, a code like: https://github.com/llogiq/mutagen/pull/43/files can't be mutated, as it throws the following error:
```
error[E0277]: the trait bound `[closure@tests/binops.rs:9:18: 30:16]: std::cmp::PartialOrd` is not satisfied
 --> tests/binops.rs:9:18
  |
9 | #[cfg_attr(test, mutate)]
  |                  ^^^^^^^ can't compare `[closure@tests/binops.rs:9:18: 30:16]` with `[closure@tests/binops.rs:9:18: 30:16]`
  |
  = help: the trait `std::cmp::PartialOrd` is not implemented for `[closure@tests/binops.rs:9:18: 30:16]`
  = note: required by `mutagen::gt`

error[E0308]: mismatched types
  --> tests/binops.rs:9:18
   |
9  |   #[cfg_attr(test, mutate)]
   |  __________________^
10 | | #[allow(unused)]
11 | | fn mutated_function() {
12 | |     let ord = 0x60;
...  |
29 | |
30 | |     if (ord < 2) {
   | |___________^ expected closure, found a different closure
   |
   = note: expected type `[closure@tests/binops.rs:9:18: 30:16]`
              found type `[closure@tests/binops.rs:9:18: 30:12 ord:_]`
   = note: no two closures, even if identical, have the same type
   = help: consider boxing your closure and/or using it as a trait object

error: aborting due to 2 previous errors

error: Could not compile `mutagen-plugin`.
```

If we should not use closures here, we shouldn't use them on eq/ne or any of the other mutations, no? What I'm missing?

llogiq:
Yes. Consider the example: `x == if foo { y } else { return }`, which with closures would become `mutagen::eq(|| x, || { if foo { y } else { return } }, $n)`. Note that the `return` now belongs to the closure, leading to a type mismatch.

llogiq:
I merged this anyway, because it unblocks another PR. Thank you!

PR #44, opened 2018-03-05
Fix mutation descriptions

gnieto:
Easy one: labels of the mutations are not in sync with the code.


plugin/src/lib.rs
```diff
@@ -355,8 +355,8 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
                             &mut self.current_count,
                             expr.span,
                             &[
-                                "replacing _ == _ with false",
```
gnieto: See: https://github.com/llogiq/mutagen/blob/master/src/lib.rs#L130

llogiq:
Thank you!

PR #45, opened 2018-03-06
Recursive mutation

manaskarekar:
Fix issue referenced in #40 and #32.

### DIFF
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -319,8 +319,8 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
                             ],
                         );
                     }
-                    let left = left.map(|e| fold::noop_fold_expr(e, self));
-                    let right = right.map(|e| fold::noop_fold_expr(e, self));
+                    let left = left.map(|e| self.fold_expr(P(e)));
+                    let right = right.map(|e| self.fold_expr(P(e)));
                     quote_expr!(self.cx, mutagen::and(|| $left, || $right, $n))
                 }
                 BinOpKind::Or => {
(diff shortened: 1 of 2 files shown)


manaskarekar:
Sorry about the second PR.

llogiq:
I had to merge another PR now, so we got a merge conflict. Can you resolve the conflict or do you need help with that? Sorry for the inconvenience.

manaskarekar:
I've resolved the merge conflicts. Please let me know if something doesn't look right.

`cargo +nightly test` on plugins does give me 'mismatched types' error at the mutate attribute but I believe that is being addressed in another issue (I haven't pulled from master recently).

plugin/src/lib.rs
```diff
@@ -319,9 +319,9 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
                             ],
                         );
                     }
-                    let left = left.map(|e| fold::noop_fold_expr(e, self));
-                    let right = right.map(|e| fold::noop_fold_expr(e, self));
-                    quote_expr!(self.cx, ::mutagen::and(|| $left, || $right, $n))
```
hmvp: Am I correct when I say that you should not remove the `::` in the last line? I think that was actually added in another commit recently

plugin/tests/binops.rs
```diff
@@ -89,7 +93,6 @@ fn test_binop_ne() {
     assert!(checker.has_multiple(noneq_msgs, "26:9: 26:15"));
 }
 
-#[test]
```
hmvp: I fail to see why the is not a test anymore. Is this intentional?

manaskarekar: Sorry, I messed up the merge.

llogiq:
Yeah, merges are hard. Anyway, this looks good. Thank you for bearing with us! :+1:

Issue #37, opened 2018-03-04
Seeking evil code


llogiq:
We want mutagen to work on as much Rust code as possible. So to test our mutation plugin, it would be prudent to gather a range of code samples designed to make the mutator emit invalid Rust code. Then we can use [compile-test](https://github.com/laumann/compiletest-rs) to ensure our mutations stay compilable.

PR #46, opened 2018-03-06
Add compiletest support

gnieto:
Adding compiletest library and adding an example for eq and ne mutations
that are applied properly.

Related to: https://github.com/llogiq/mutagen/issues/37

On the other hand, I'm adding some tests that currently are failing (on
the compile-fail folder) that we should fix at some time. I think it
may be useful to keep track of the issues pending to be fixed and it
will help to not add regressions of known patterns that caused issues on
the past.

@llogiq @hmvp I'm not sure about keeping track of the "compile-fail" tests. What do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,4 +12,4 @@ Cargo.lock
 **/*.rs.bk
 
 # Generated file from integration tests
-/plugin/tests/target/
+**/mutations.txt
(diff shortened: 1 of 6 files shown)


llogiq:
Thanks! I think it's good to keep track of errors, so we can see improvement and also to document limitations.

PR #47, opened 2018-03-08
Pass through cargo test args from the runner

llogiq:
This allows us to e.g. compile with `--release` or `--features`.

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -1,6 +1,5 @@
 extern crate json;
 
-
 use std::process::{Command, Stdio};
 use std::fs::File;
 use std::io::Read;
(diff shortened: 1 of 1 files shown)


runner/src/main.rs
```diff
@@ -100,6 +100,7 @@ fn get_mutations_filename() -> PathBuf {
 fn compile_tests() -> PathBuf {
     let compile_out = Command::new("cargo")
         .args(&["test", "--no-run", "--message-format=json"])
+	.args(std::env::args_os())
```
gnieto: This indentation looks weird

llogiq: Good catch!

PR #48, opened 2018-03-08
simplify mutagen lib

llogiq:
This pushes most od the code outside the `Mutagen` object and adds a `reset` method to ease testing.

I'm unsure if we should expose this method (or `next()`), but it currently seems the best way forward.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -54,224 +54,175 @@ pub struct Mutagen {
 
 impl Mutagen {
     /// get the current mutation count
+    #[inline]
     pub fn get(&self) -> usize {
         self.x.load(Ordering::Relaxed)
     }
 
-    /// check if the argument matches the current mutation count
-    ///
-    /// this simplifies operations like `if mutagen::MU.now(42) { .. }`
-    pub fn now(&self, n: usize) -> bool {
-        self.get() == n
-    }
-
     /// increment the mutation count
+    #[inline]
     pub fn next(&self) {
         //TODO: clamp at maximum, return Option<usize>, impl Iterator?
         self.x.fetch_add(1, Ordering::SeqCst);
     }
 
-    /// insert the original or an alternate value, e.g. `MU.select(&[2, 0], 42)`
-    pub fn select<T, S: Selector<T>>(&self, selector: S, n: usize) -> T {
-        selector.get(self.get().wrapping_sub(n))
-    }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -280,45 +231,43 @@ mod tests {
 
     #[test]
     fn eq_mutation() {
-        let mu = Mutagen { x: AtomicUsize::new(0) };
-
+        reset();
```
gnieto: I think we can get rid of `reset` and `next` using:
```
MU.x.store(n, Ordering::SeqCst)
```

llogiq:
Argh. Unfortunately, tests can be run in parallel. This is unfortunate, as it precludes doctests. Then again, we probably shouldn't worry about them anyway, and just hide all but the crate docs.

src/lib.rs
```diff
@@ -280,45 +231,43 @@ mod tests {
 
     #[test]
     fn eq_mutation() {
```
gnieto: if we remove `next` and `reset` methods, we can rewrite the test as:

```rust
        let n = 1;
        MU.x.store(1, Ordering::SeqCst);
        assert_eq!(true, eq(|| 1, || 0, n));

        // Always false
        MU.x.store(2, Ordering::SeqCst);
        assert_eq!(false, eq(|| 1, || 0, n));

        // Checks inequality
        MU.x.store(3, Ordering::SeqCst);
        assert_eq!(true, eq(|| 0, || 1, n));
        assert_eq!(false, eq(|| 1, || 1, n));

        // Checks equality
        MU.x.store(4, Ordering::SeqCst);
        assert_eq!(true, eq(|| 0, || 0, n));
        assert_eq!(false, eq(|| 1, || 0, n));
```

src/lib.rs
```diff
@@ -280,45 +231,43 @@ mod tests {
...
-        let mu = Mutagen { x: AtomicUsize::new(0) };
```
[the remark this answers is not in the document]
llogiq: Exactly. I'll roll this back.

src/lib.rs
llogiq: Possibly, but `reset` and `next` make the meaning more clear, I think.

llogiq:
I'll rather start from a clean slate and keep the tests being able to run in parallel.

Issue #21, opened 2018-02-19
Mutation: Clone values to other parts of the function


hmvp:
Instead of using default to generate different values for assignments or arguments we can clone values from different places in the same function and use them instead

```rust
fn times_two(arg: u8) -> u8 {
     arg * 2
}
```
mutate to
```rust
fn times_two(arg: u8) -> u8 {
     // Original value as head, all other similar typed values as tail
     mutagen::clone(&[arg * 2, arg], n)
}
```
with:
```rust
fn clone<T>(choices: &[T], n: usize ) -> T {
   let x = n.wrapper_sub(self.get());

   let x = if x < 0 && x >= choices.len() {0} else { x };

   choices[x].clone()
}
```

llogiq:
Wouldn't that at least require `T: Clone`? Also we already do early return with arguments, so a part of this is already taken care of.

llogiq:
I think I have a solution for this and a few other mutations. Will write something up soon.

llogiq:
See https://llogiq.github.io/2018/03/03/opportune.html

Issue #12, opened 2018-02-18
Mutation: Swap similar operators


hmvp:
One option would to swap + and - or * and %. 

Also similar for `x += 1` and `x -=1` etc

However this will probably trigger under/overflow issues.

llogiq:
This can be done with specialization so we let typeck infer the correct code for us.

PR #49, opened 2018-03-08
harden plugin

### LINKED ISSUE #34: && , || must evaluate their first, == , != both arguments
In this example:

```rust
fn foo() -> usize { 2 }
fn bar() -> usize { 23 }

fn blubb() -> usize {
    let mut value;
    if ({ value = foo(); value > 5 }) || ({ value = bar(); value < 5 }) {
        value + 1
    } else {
        42
    }
}

fn main() {
    println!("{}", blubb());
}
```

We cannot mutate the `||` to `true` or `false`, because then `value` would not be set, so the compiler will bail. This also means we cannot in general rely on closures to defer actions, because of the side effects / returns we might lose. Solving this for `||` and `&&` is perhaps not trivial, but workable:

For example, `x || y` can be mutated to:

```rust
(match ($x, mutagen::diff($n)) {
    (_, 0) => false,
    (_, 1) => true,
    (x, 2) => x,
    (x, 3) => !x,
    (x, n) => x || ($y) == (n != 4),
})
```

llogiq:
This fixes #34 and removes unneeded closures for all operations. With the last commit, we also get a first step towards #21 and #12, among others. 

I also ran rustfmt over the plugin.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,3 +1,4 @@
+#![feature(specialization)]
 /// Welcome to the mutagen crate. Your entry point will probably be cargo-mutagen, so install it
 /// right away.
 #[macro_use]
@@ -5,6 +6,9 @@ extern crate lazy_static;
 
 use std::env;
 use std::sync::atomic::{AtomicUsize, Ordering};
+
+mod ops;
+pub use ops::*;
 
 /// A helper trait to select a value from a same-typed tuple
 #[doc(hidden)]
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -65,131 +72,86 @@ impl Mutagen {
...
-                let yv = y();
-
-                xv < yv || xv > yv
-            }, // != with PartialOrd
-            _ => x() > y()
+            2 => x < y,
+            3 => x <= y,
+            4 => x >= y,
+            5 => x == y,
+            6 => x != y, // != with PartialOrd
```
gnieto: I think the comment is not needed anymore

llogiq:
Thanks for the review! While I was at it, I added a build script to create some traits for ops.

PR #52, opened 2018-03-09
This closes #16

### LINKED ISSUE #16: Mutation: Switch arguments of same type
The plugin already has the analysis. There are but two wrinkles to take care of:

* [ ] It may make a difference if the args are taken by ref or by value, the analysis pass doesn't take that into account yet
* [ ] we should take care not to issue the same switch twice. This could probably be done easily by filtering the argument pairs so only pairs where the first name compares lower than the last one are switched.

rajcspsg:
This is PR for issue https://github.com/llogiq/mutagen/issues/16

### DIFF
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -724,8 +724,8 @@ fn fold_first_block(block: P<Block>, m: &mut MutatorPlugin) -> P<Block> {
                         .unwrap(),
                 );
             }
-            for (ref key, ref values) in (&interchangeables).iter() {
-                for value in (&values).iter() {
+            for (ref key, ref values) in interchangeables {
+                for value in values.iter() {
                     let n = *current_count;
                     let key_ident = key.to_ident();
                     let value_ident = value.to_ident();
@@ -734,13 +734,7 @@ fn fold_first_block(block: P<Block>, m: &mut MutatorPlugin) -> P<Block> {
                         mutations,
                         current_count,
                         block.span,
-                        &[
-                            &format!(
-                                "exchange {} with {}",
-                                key_ident.as_str(),
-                                value_ident.as_str()
-                            ),
-                        ],
+                        &[&format!("exchange {} with {}", key.as_str(), value_ident)],
                     );
                     pre_stmts.push(
                         quote_stmt!(cx,
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -116,13 +116,13 @@ impl<'a, 'cx> MutatorPlugin<'a, 'cx> {
         // arguments of output type
         let mut have_output_type = vec![];
         // add arguments of same type, so we can switch them?
-        let mut argtypes: HashMap<Symbol, (Mutability, &Ty)> = HashMap::new();
-        let mut typeargs: HashMap<(Mutability, &Ty), Vec<Symbol>> = HashMap::new();
+        let mut argtypes: HashMap<Symbol, (BindingMode, &Ty)> = HashMap::new();
+        let mut typeargs: HashMap<(BindingMode, &Ty), Vec<Symbol>> = HashMap::new();
         for arg in &decl.inputs {
-            if let Some((name, mutability)) = get_pat_name_mut(&arg.pat) {
-                argtypes.insert(name, (mutability, &*arg.ty));
+            if let Some((name, binding_mode)) = get_pat_name_mut(&arg.pat) {
+                argtypes.insert(name, (binding_mode, &*arg.ty));
                 typeargs
-                    .entry((mutability, &arg.ty))
+                    .entry((binding_mode, &arg.ty))
                     .or_insert(vec![])
                     .push(name);
                 if Some(&*arg.ty) == out_ty {
@@ -137,8 +137,8 @@ impl<'a, 'cx> MutatorPlugin<'a, 'cx> {
                     arg,
                     typeargs[&mut_ty]
                         .iter()
+                        .filter(|a| a >= &&arg)
                         .cloned()
-                        .filter(|a| a != &arg)
                         .collect(),
                 );
             }
@@ -724,6 +724,33 @@ fn fold_first_block(block: P<Block>, m: &mut MutatorPlugin) -> P<Block> {
                         .unwrap(),
                 );
             }
+            for (ref key, ref values) in (&interchangeables).iter() {
+                for value in (&values).iter() {
+                    let n = *current_count;
+                    let key_ident = key.to_ident();
+                    let value_ident = value.to_ident();
+                    add_mutations(
+                        cx,
+                        mutations,
+                        current_count,
+                        block.span,
+                        &[
+                            &format!(
+                                "exchange {} with {}",
+                                key_ident.as_str(),
+                                value_ident.as_str()
+                            ),
+                        ],
+                    );
+                    pre_stmts.push(
+                        quote_stmt!(cx,
+                        if ::mutagen::now($n) { 
+                            let ($key_ident, $value_ident) = ($value_ident, $key_ident)
+                         }).unwrap(),
+                    );
+                }
+            }
+            //let ($a, $b) = if mutagen::now($n) { ($b, $a) } else { ($a, $b) };
             //TODO: switch interchangeables, need mutability info, too
             //for name in method_info.interchangeables { }
         }
@@ -768,14 +795,9 @@ fn add_mutations(
     *count += descriptions.len();
 }
 
-fn get_pat_name_mut(pat: &Pat) -> Option<(Symbol, Mutability)> {
+fn get_pat_name_mut(pat: &Pat) -> Option<(Symbol, BindingMode)> {
     if let PatKind::Ident(mode, i, _) = pat.node {
-        Some((
-            i.node.name,
-            match mode {
-                BindingMode::ByRef(m) | BindingMode::ByValue(m) => m,
-            },
-        ))
+        Some((i.node.name, mode))
     } else {
         None
     }


plugin/src/lib.rs
```diff
@@ -724,6 +724,7 @@ fn fold_first_block(block: P<Block>, m: &mut MutatorPlugin) -> P<Block> {
                         .unwrap(),
                 );
             }
+            //let ($a, $b) = if mutagen::now($n) { ($b, $a) } else { ($a, $b) };
             //TODO: switch interchangeables, need mutability info, too
             //for name in method_info.interchangeables { }
```
llogiq: You should uncomment this loop. The `name' is wrong, we need `(ref key, ref values)` pairs, then iterate over `values`. Now in that inner loop we have two `Symbol`s we want to switch (to use them wihin a `quote!(..)`, we need to get their Identifier, so call `.to_ident()` as above in line 713.

Next, we need to add the mutation to our mutation list (so add_mutations(..) with a single mutation. The text should be `"switch the {key} and {value} argument"` (format! with key and value). See lines 712 and following. Finally, we need to push our mutation statement, which we can get by `quote!`ing the code in line 727's comment (just change `$a` and `$b` as necessary).

llogiq:
You're almost there! :+1: See my comment above. Again, if you need anything else, don't hesitate to ask.

Tip: if you write "This closes #16" in the PR description, the issue will automatically close on merge.

plugin/src/lib.rs
```diff
@@ -724,6 +724,18 @@ fn fold_first_block(block: P<Block>, m: &mut MutatorPlugin) -> P<Block> {
                         .unwrap(),
                 );
             }
+            let (ref key, ref values) = for method_info in info.method_infos {
+                for interchangeable in method_info.interchangeables {
+                    let ident = interchangeable.name.to_ident();
```
rajcspsg: @llogiq - I still need your help. When I try to compile the code I'm getting error.

```

//let ($a, $b) = if mutagen::now($n) { ($b, $a) } else { ($a, $b) };
             //TODO: switch interchangeables, need mutability info, too
             //for name in method_info.interchangeables { }

```

Do I've to move the for loop inside the if [](https://github.com/llogiq/mutagen/blob/master/plugin/src/lib.rs#L695)

plugin/src/lib.rs
```diff
@@ -724,6 +724,23 @@ fn fold_first_block(block: P<Block>, m: &mut MutatorPlugin) -> P<Block> {
                         .unwrap(),
                 );
             }
+            let (ref key, ref values) = for (key, &value) in interchangeables.iter() {
+                let n = *current_count;
+                let ident = key.to_ident();
+                add_mutations(
+                    cx,
+                    mutations,
```
rajcspsg: @llogiq - Is what I'm doing here, correct? I'm not sure how to call `add_mutations` with single mutation.

plugin/src/lib.rs
```diff
@@ -724,6 +724,23 @@ fn fold_first_block(block: P<Block>, m: &mut MutatorPlugin) -> P<Block> {
...
+                    current_count,
+                    block.span,
+                    &["insert return default()"],
+                );
+                if mutagen::now(n) {
+                    (value, key)
+                } else {
+                    (key, value)
+                };
```
rajcspsg: I'm getting here an error

```
if and else have incompatible types
   --> src/lib.rs:737:17
    |
737 | /                 if ::mutagen::now(n) {
738 | |                     (value, key)
739 | |                 } else {
740 | |                     (key, value)
741 | |                 };
    | |_________________^ expected struct `std::vec::Vec`, found reference
```

plugin/src/lib.rs
```diff
@@ -724,6 +724,23 @@ fn fold_first_block(block: P<Block>, m: &mut MutatorPlugin) -> P<Block> {
                         .unwrap(),
                 );
             }
+            let (ref key, ref values) = for (key, &value) in interchangeables.iter() {
```
llogiq: that line should probably read `for (ref key, ref values) in &interchangesables {`.
Please note that this means key is a `&Symbol` and values is a `&Vec<Symbol>`, so we also need an inner `for value in values {` loop here.

plugin/src/lib.rs
llogiq: Almost – the message should read `&format!("exchange {} with {}", ident.as_str(), value_ident.as_str())`. For that you also need to get the ident of the value (same as with the key).

plugin/src/lib.rs
llogiq: Yes, you need this within a `quote!`(same as with the other mutations). So wrap it in `quote!(cx, /*YOUR CODE HERE*/)` and change the `n` to `$n` (this gets our mutation number into the generated code), and change `key` and `value` to `$key_ident` and `$value_ident` respectively. Now you only need to prepend the `let ($key_ident, $value_ident) = ` before that `if` so the mutation will actually work.

plugin/src/lib.rs
```diff
@@ -724,8 +724,8 @@ fn fold_first_block(block: P<Block>, m: &mut MutatorPlugin) -> P<Block> {
                         .unwrap(),
                 );
             }
-            for (ref key, ref values) in (&interchangeables).iter() {
-                for value in (&values).iter() {
+            for (ref key, ref values) in interchangeables {
+                for value in values.iter() {
```
rajcspsg: @llogiq - does the for loop that I implemented here looks correct?

plugin/src/lib.rs
```diff
@@ -724,6 +724,33 @@ fn fold_first_block(block: P<Block>, m: &mut MutatorPlugin) -> P<Block> {
...
+                    add_mutations(
+                        cx,
+                        mutations,
+                        current_count,
+                        block.span,
+                        &[
+                            &format!(
+                                "exchange {} with {}",
+                                key_ident.as_str(),
+                                value_ident.as_str()
```
rajcspsg: I get an error during compilation. How can I resolve this error?
```
error[E0599]: no method named `as_str` found for type `syntax::ast::Ident` in the current scope
   --> src/lib.rs:740:43
    |
740 |                                 key_ident.as_str(),
    |                                           ^^^^^^

error[E0599]: no method named `as_str` found for type `syntax::ast::Ident` in the current scope
   --> src/lib.rs:741:45
    |
741 |                                 value_ident.as_str()
    |                                             ^^^^^^
```

gnieto: I think you can write:

```rust
                                key.as_str(),
                                value.as_str()
```

The reason is that `key_ident` is of type `Ident`, which does not have the `as_str` method. On the other hand, `key` is of type `Symbol`, which has the `as_str` method.

rajcspsg: Thanks @gnieto .

I've changed the code as below and yet to update it.


```
for (ref key, ref values) in interchangeables {
                for value in values.iter() {
                    let n = *current_count;
                    let key_ident = key.to_ident();
                    let value_ident = value.to_ident();
                    add_mutations(
                        cx,
                        mutations,
                        current_count,
                        block.span,
                        &[&format!("exchange {} with {}", key.as_str(), value_ident)],
                    );
                    pre_stmts.push(
                        quote_stmt!(cx,
                        if ::mutagen::now($n) { 
                            let ($key_ident, $value_ident) = ($value_ident, $key_ident)
                         }).unwrap(),
                    );
                }
            }
```

does it looks good?

gnieto: Looks good to me

rajcspsg: Thanks @gnieto . Currentlt tests in `plugin/` are failing but not due to this PR as tests are failing in master branch too. 
I will wait for @llogiq to approve this PR

plugin/src/lib.rs
llogiq: A `&` before `interchangeables` will help here, as we don't need to consume it. This will use `&Vec`s `IntoIterator` implemenation, so you don't need to call `.iter()` (which is another way to do the same thing)

rajcspsg: @llogiq - I tried that too `for (ref key, ref values) in &interchangeables {` but got compilation below error.

```

error[E0277]: the trait bound `&&std::collections::HashMap<syntax::ast::Symbol, std::vec::Vec<syntax::ast::Symbol>>: std::iter::Iterator` is not satisfied
   --> src/lib.rs:728:42
    |
728 |             for (ref key, ref values) in &interchangeables {
    |                                          ^^^^^^^^^^^^^^^^^ `&&std::collections::HashMap<syntax::ast::Symbol, std::vec::Vec<syntax::ast::Symbol>>` is not an iterator; maybe try calling `.iter()` or a similar method
    |
    = help: the trait `std::iter::Iterator` is not implemented for `&&std::collections::HashMap<syntax::ast::Symbol, std::vec::Vec<syntax::ast::Symbol>>`
    = note: required by `std::iter::IntoIterator::into_iter`
```

Could you please suggest some ideas how to resolve this error.

llogiq: Ah, my mistake: `interchangeables` already is a reference. In that case, you don't need the `&`.

llogiq:
Thank you!

rajcspsg:
@llogiq - Thanks a lot for accepting my PR. It is very pleasure to working with you and answering all my silly questions. In fact it is my first PR that I've don't any coding at all :)

llogiq:
@rajcspsg I'll be happy to mentor you further, if you want to pick up another issue. Again sorry if I make mistakes every now and then. I have to juggle a job and three kids, so I may sometimes lose the big picture.

rajcspsg:
@llogiq - No problem. I will pick another beginner issue and I will comment on the issue itself :+1:

PR #56, opened 2018-03-12
fix some warnings

rleungx:


llogiq:
Thank you! :+1: In general, we might want to warn if removing files fails, but this can be done in another PR.

PR #57, opened 2018-03-12
support more than one executable

### LINKED ISSUE #55: Runner - Support for testsuites with more than one executable
At the moment, the runner finds only one test executable.

The method `compile_tests` compiles the test and returns a single `PathBuf`, but it should return a vector of them, and execute each of the found test binaries. Note that we are early-returning when we find the first path.

With this change, mutagen will be able to run on projects which has both unit, integration and doc tests.

rleungx:
This pull request partly resolves #55 since doctests can not work this way.

### DIFF
--- a/runner/src/main.rs
+++ b/runner/src/main.rs
@@ -85,7 +85,8 @@ fn get_mutations_filename() -> PathBuf {
     mutagen_dir.join(MUTATIONS_LIST)
 }
 
-fn compile_tests() -> PathBuf {
+fn compile_tests() -> Vec<PathBuf> {
+    let mut tests: Vec<PathBuf> = Vec::new();
     let compile_out = Command::new("cargo")
         .args(&["test", "--no-run", "--message-format=json"])
         .args(std::env::args_os())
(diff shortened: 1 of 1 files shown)


gnieto:
This solves the issue as described, but:

I reported the issue, and I've just realized that there's an issue. Let me explain: Mutagen plug-in is applied when a test is compiled and it generates a file on `target/mutagen/mutations.txt`. When we run `cargo test --no-run`, we are generating that file. But the issue is that, as we compile all the tests (lib, examples, ...) at once, it will overwrite the mentioned file (`target/mutagen/mutations.txt`) when compiling each test suite.

I don't know how we can solve this limitation; maybe @llogiq has some idea on how we can work around this limitation.

llogiq:
Interesting. I wasn't aware of this behavior. In any event, is that a problem? We shouldn't mutate test code anyway, only the code under test, and that code should basically stay the same every time. So writing the same `mutations.txt` file three times may be wasteful, but not problematic, unless there are mutations in more than one compiled crate, or am I missing something?

llogiq:
Working around this limitation would involve deleting the `mutations.txt` file at the start of the test runner, and then either unconditionally appending and removing dupes later, or changing the format to a serialized map by span + mutation type to mutation number + description so we can do lookup.

gnieto:
hey @llogiq . I think you are ok and this should not be a problem. Sorry for the false alarm

llogiq:
No problem. I've now thought about it some more, and I agree we should try to generate multiple mutations.txt files and see if they differ.

llogiq:
Thank you @rleungx! We can experiment later regarding the multiple files.

PR #59, opened 2018-03-13
orderly destructure patterns

llogiq:
Until this PR, mutagen just ignored anything but plain ident patterns, so if you have `fn foo((a, b): (A, B))`, you were out of luck. This PR destructures all bindings and tries to keep as much information as possible.


llogiq:
Ah, I forgot that types don't implement equality correctly – they also compare NodeIds. So we either need to transform them to our own type or create a wrapper or comparison function. Clippy already has something like this.

Good idea with the wrapper type! I'll extend my code.

PR #60, opened 2018-03-15
add assign ops mutation traits

llogiq:
I'm going to try to refactor some more before I get the plugin to use the traits, but I wanted to introduce the corresponding assign ops first.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -29,11 +29,33 @@ where T: {0}<Rhs>,
         }}
     }}
 }}
+
+pub trait {0}{2}Assign<Rhs=Self> {{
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -29,11 +29,33 @@ where T: {0}<Rhs>,
...
+            {0}Assign::{1}_assign(self, rhs);
+        }}
+    }}
+}}
 ", o_trait, o_fn, mut_trait, mut_fn)
 }
 
 static BINOP_PAIRS: &[[&str; 4]] = &[
-    ["Add", "add", "Sub", "sub"],
+["Add", "add", "Sub", "sub"],
```
hmvp: I assume this is unintentional. You might want to fix this. I don't mind if you don't

PR #61, opened 2018-03-16
Fix exchange mutation

gnieto:
There were some issues with the current code:

- Hash does not work as expected (proposed a solution, but I'm aware that
  probably we need a better solution)
- Missing semi-colon at then end of the `quote_stmt`
- Combination of all possible interchanges was not working as expected,
  and it issued some mutations that were repeated

Finally, I added some tests to ensure that we are generating exchange mutations.


plugin/src/lib.rs
```diff
@@ -106,6 +107,15 @@ impl<'t> PartialEq for ArgTy<'t> {
     }
 }
 
+impl<'t> Hash for ArgTy<'t> {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.0.hash(state);
+        let s = format!("{:?}", self.1.node);
```
gnieto: Ok, I agree that this is not elegant at all, but it works. Are you aware of any work around? (Note that, previous to this change, the `typeargs` hash is not working as expected, and it is creating a new entry for any kind of type that is pushed, even if it is the same type)

plugin/tests/binops.rs
```diff
@@ -1,128 +0,0 @@
-#![cfg_attr(test, feature(plugin))]
```
gnieto: This has been renamed to `integration.rs`. As tests are compiled and then ran, `target/mutagen/mutations.txt` is overwritten, so we can't have multiple test files.
I'm working on a branch that extends current `compiletest`-tests that would allow to have multiple test files.

plugin/src/lib.rs
llogiq: Agreed that it is not elegant, but wie can fix that later. What worries me more is that our `PartialEq` is either not correct w.r.t. lifetimes or violates the trait's reflexivity requirement. To fix this, we'll need to add the argument position and set `inout` to `a.pos == b.pos`.

llogiq:
Looks good to me, modulo the failing tests.

llogiq:
Thanks! Good to have this working. I'll see if I can improve the hash function.

PR #63, opened 2018-03-19
use failure for error handling

### LINKED ISSUE #58: Runner: error handling with failure
I think our runner could benefit from doing error handling with the [failure](https://crates.io/crates/failure) crate. Using boxed errors should be an acceptable compromise, and should get rid of all `unwrap`s and `expect`s.

The binary should on error simply write out the error and exit with status=1, otherwise just exit normally.

rleungx:
Resolves #58.


llogiq:
Thank you! :+1:

PR #66, opened 2018-03-20
 Avoid mutate expressions on match Range patterns 

gnieto:
Match may have distinct types of patterns. When they are numeric or char
ranges, they need to be literals, but not expressions. If we do that mutation, we get the following error:

```
error[E0029]: only char and numeric types are allowed in range patterns
```

Now, when we detect a `Pat` which is of `PatKind::Range`, we are **not**
recursively mutating expressions. Otherwise, we are mutating as we were
doing before.


llogiq:
Good catch! We may however want to change our mutation to change the literal value instead of adding/subtracting `1` in the future, which would then also work with range patterns. I'll add an issue.

PR #68, opened 2018-03-25
Change coverage flags to use a per-fn [AtomicUsize;_ ]

llogiq:
This is a thing I wanted to do for some time, and I finally got around to doing it :slightly_smiling_face:. The solution here is to first create a `static <genkey#N>: [AtomicUsize; 0]`, then do all the mutations, come back and patch up the length.

### DIFF
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -5,10 +5,11 @@ extern crate syntax;
 
 use rustc_plugin::registry::Registry;
 use std::collections::HashMap;
-use std::convert::TryFrom;
+use std::convert::{TryFrom, TryInto};
 use std::fs::{create_dir_all, File, OpenOptions};
 use std::hash::{Hash, Hasher};
 use std::io::{BufWriter, Write};
+use std::mem;
 use std::sync::atomic::{AtomicUsize, Ordering::SeqCst};
 use syntax::ast::*;
 use syntax::codemap::{Span, Spanned};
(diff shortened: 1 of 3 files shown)


plugin/src/lib.rs
```diff
@@ -397,10 +437,20 @@ impl<'a, 'cx> MutatorPlugin<'a, 'cx> {
                     ],
                 );
                 quote_expr!(self.cx(), {
-                    ::mutagen::report_coverage($n...$current);
+                    ::mutagen::report_coverage($n..$current, &$sym[$flag], $mask);
                     ::mutagen::ge($right, $left, $n)
                 })
             }
+            BinOpKind::Add => {
```
gnieto: Adding this new mutation is intentional?

src/coverage.rs
```diff
@@ -2,41 +2,28 @@ use std::env;
 use std::fs::OpenOptions;
 use std::io::Write;
 use std::ops::Range;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicUsize, Ordering};
 
-lazy_static! {
-    static ref COVERAGE_ALREADY_CHECKED: Vec<AtomicBool> = {
-        let max = env::var("MUTAGEN_MUTATION_AMOUNT").map(|s|s.parse().unwrap_or(0)).unwrap_or(0) + 1;
```
gnieto: On the runner, we are still passing this envvar. We should remove this from the runner code.

plugin/src/lib.rs
llogiq: Yes, I wanted to see if it works with opportunistic mutations. I'm going to extend the build script to generate code for all of them to be included in this place next.

src/coverage.rs
llogiq: Agreed. :+1:

PR #69, opened 2018-03-26
UI-like tests for mutation plugin

gnieto:
I've been working on this, even if i'm not sure if it's a good idea: I think it's a better approach of the fixed and hand-crafted tests that we had before, but it also comes with some drawbacks.

I've mixed feeling about this change, so feel free to close the PR!

:heavy_check_mark: They much more are easy and straight forward to write (it's just write an example)
:heavy_check_mark: Old tests and specific code for tests which was parsing the mutations file can be removed
:small_red_triangle_down: Depends on the unstable feature "test"
:small_red_triangle_down: It depends on the format of `mutations.txt` (so, if it changes, tests will "break")
:small_red_triangle_down: We need to sort the mutations file to be able to compare the files (I guess that this is due to `interchangeables` depending on a `HashMap`, which probably does not guarantee order)


plugin/tests/mutagen/update-all-references.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
gnieto: I've borrowed this from clippy. I'm not sure if I can borrow it as-is.

llogiq: I think I can get a specific license for use in mutagen.

power-fungus:
Are parts of this pull-request still relevant for the current architecture? What tests are added by this PR?

gnieto:
Hi @power-fungus , I don't think this is relevant anymore. The idea was to create a set of tests, similar to clippy, to ensure that a source code file always produced the same mutation.

PR #73, opened 2018-04-03
Add mutation for iterators to skip elements

gnieto:
Added a mutation for for-loops to mutate the iterator to the following
cases:

- Empty iterator (no elements will be yielded)
- Skip first element
- Skip last element
- Skip first and last element

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,6 +9,7 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 
 mod ops;
 pub use ops::*;
+mod iterators;
 
 mod coverage;
 pub use coverage::report_coverage;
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -161,6 +162,16 @@ impl Mutagen {
             _ => x >= y,
         }
     }
+
+    pub fn forloop<'a, I: Iterator + 'a>(&self, i: I, n: usize) -> Box<Iterator<Item=I::Item> + 'a> {
```
gnieto: If I'm not wrong, I can't use the `impl Trait` notation as it's meant to be used for static dispatch (at least, I couldn't make it work).
I don't know if there's any work around to avoid boxing the iterator.

llogiq:
Thank you once more.

PR #75, opened 2018-04-03
Check runner on example project

gnieto:
Trying to run the plugin on the example project, with and without
coverage.
This should allow to avoid some regressions and make sure that we don't
break the runner.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,3 +10,9 @@ script:
   - pushd plugin
   - cargo test
   - popd
+  - pushd runner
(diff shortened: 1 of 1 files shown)


gnieto:
Depends on: https://github.com/llogiq/mutagen/pull/74

gnieto:
As https://github.com/llogiq/mutagen/pull/80 has been merged, the example project now can be mutated and the build now is ok, so it's not WIP anymore.

PR #76, opened 2018-04-03
Add some code examples that fail to compile

gnieto:


plugin/tests/compile-fail/oportunistic-issues.rs
```diff
@@ -0,0 +1,19 @@
+#![feature(plugin)]
+#![plugin(mutagen_plugin)]
+#![feature(custom_attribute)]
+extern crate mutagen;
+
+#[mutate]
+fn main() {
+    let mut count = 0;
+    count += 1; //~  mismatched types [E0308]
+}
```
llogiq: This is really an error. I think we are missing a `&mut` for the left-hand side argument.

gnieto: Yes, adding the `&mut` fixes the issue, at least, on the code I was mutating

plugin/tests/compile-fail/oportunistic-issues.rs
```diff
@@ -0,0 +1,19 @@
...
+
+#[mutate]
+//~^ the trait bound `u32: mutagen::MulDiv<i32>` is not satisfied [E0277]
+fn muldiv_issue() {
+    let val: u32 = 5;
+    let b = val * 2i32;
+}
```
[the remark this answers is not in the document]
gnieto: You are right. I got an error with the `MulDiv` trait and I tried to reduce it to the minimal example, but it seems that I failed to do it :/
I'll check again the original code where I found this and figure out if it's a real issue with the mutated code.

llogiq:
I think the problem is really the Shl / Shr traits which are implemented on far too many types.

I [have a playground](https://play.rust-lang.org/?gist=bbdf3edc210b03168692951332428281&version=nightly) with everything but the left shift "mutated" (for the sake of typeck), and it works.

PR #81, opened 2018-04-09
disable shift mutation

llogiq:
As long as we haven't fixed the problem with Shl/Shr forwarding, let's disable mutating bit shifts for now.


llogiq:
This is no longer needed, as the shift mutations are now fixed.

PR #82, opened 2018-04-09
Improve runner output

gnieto:
Improve a little bit mutagen's runner:

- Added colors for better visibility.
- Remove tests output when tests are ok; as we count non-failing tests
  as failure, the output was always the same: All tests green.
- Avoid saving the output of the test commands.

### DIFF
--- a/runner/Cargo.toml
+++ b/runner/Cargo.toml
@@ -11,6 +11,7 @@ keywords = ["testing"]
 [dependencies]
 json = "0.11.12"
 failure = "0.1.1"
+colored = "1.6"
 
 [badges]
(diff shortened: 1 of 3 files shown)


llogiq:
I'm not against colors per se, but as it is, this won't work on Windows, right?

gnieto:
@llogiq You are right, it won't work on Windows. I've removed that part, what do you think about removing the tests output when they are ok? With enough tests, it gets verbose and I think it's not useful.

llogiq:
Agreed, lets nix the test output.

llogiq:
We can add color by flag (it might be useful to see how rustc handles this) later.

PR #84, opened 2018-04-12
Add bounded iterations for loop loops

### LINKED ISSUE #23: Bounded iteration
Apart from timeouts (which aren't implemented yet either) which the mutation test runner needs to take care of, we should probably introduce bounds for all loops we encounter. This means we issue a loop ID for each iteration, and modify each loop to setup a "loop run" object that records the number of iterations (for the baseline test run, i.e. mutation_count == 0). On drop, the loop run object compares its iteration count with the maximum loop count for the loop ID and if higher, updates the maximum loop count.

Then on mutation runs, the object compares the current loop count with *a base value* `+` *the recorded maximum* `*` *some factor* and if higher, panics, thus limiting test runtime without needing to kill the process (on the other hand, we may want to kill it nevertheless, because a failing test may leave the process in a detrimental state). Even then, we allocate less runtime for a test that is very likely to fail.

gnieto:
This closes #23 .

It introduces a mutation only for `loop` loops which instruments it and traces the max bound for that loop (if mutation == 0) or it imposes a maximum bound (if mutation > 0; now it's fixed to 1000 iterations).

Once `loop`s will be finished, we can propagate this to the other kind of loops.


llogiq:
Good idea to move out bounded loops into their own plugin.

gnieto:
I've been a little bit busy this last days. I'll try to retake the work on this soon.

src/bounded_loop.rs
```diff
@@ -0,0 +1,156 @@
...
+use std::fs::OpenOptions;
+use std::io::Write;
+use std::io::{BufReader, BufRead, Read};
+use std::fs::File;
+
+lazy_static!{
+    static ref MUTAGEN_LOOP_COUNT: HashMap<LoopId, usize> = {
+        let mut h = HashMap::new();
+
+        File::open("target/mutagen/loops.txt").map(|mut file| {
```
gnieto: I was thinking about doing some abstraction for IPC, as there's some repeated code with the coverage communication, but I postponed this, as I wanted to finish this task first.

src/bounded_loop.rs
```diff
@@ -0,0 +1,156 @@
...
+            let reader = BufReader::new(file);
+
+            for l in reader.lines() {
+                let line = l.unwrap_or(String::from(""));
+                let splits: Vec<&str> = line.split(",").collect();
```
gnieto: I was thinking on adding `serde` and  `serde_json` as depdency. On one hand, I think that maybe it's too much, but on the other it's nicer to use. Do you have any preference?

src/bounded_loop.rs
```diff
@@ -0,0 +1,156 @@
...
+impl<'a> LoopStep for LoopCount<'a> {
+    /// step is called on each iteration of a loop
+    fn step(&mut self) {
+        self.count = self.count.saturating_add(1);
+    }
+}
+
+impl<'a> Drop for LoopCount<'a> {
+    fn drop(&mut self) {
+        // TODO: llogiq suggested to use atomic max as the following code is not accurate due to TOCTOU
```
gnieto: If I'm not wrong, on some place you talked about using an atomic function which saves the max, but I didn't found it on the documentation. Maybe I'm not looking on the right place or my memory fails.

src/bounded_loop.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+
+impl LoopStep for LoopBound {
+    /// step is called on each iteration of a loop
+    fn step(&mut self) {
+        self.count = self.count.saturating_add(1);
+
+        if self.bound == self.count {
+            ::std::process::exit(-2);
```
gnieto: I did this planning to show an special status on the runner, but I'm not sure it's worth to do it.

src/bounded_loop.rs
llogiq: `AtomicUsize::fetch_max`. I added it a few weeks ago, and it's unstable, so it's only in the nightly docs.

src/bounded_loop.rs
llogiq: I think [men_file](https://docs.rs/men_file) could work well for us.

gnieto:
- Added the other types of loop (`While`, `WhileLet` and `ForLoop`)
- Use `fetch_max` to keep track of the loop bounds

Can you check it now? Do you think there's something missing?

src/bounded_loop.rs
llogiq: I also think it's a bit much. All we need here are <loop ID>, <count> pairs. Also a clean mutagen build is already in the 100s of megabytes, so I'm a bit wary of adding such a large dependency.

src/bounded_loop.rs
llogiq: I like the idea of being able to identify loop overflows from the runner, this would enable us to do better reporting.

plugin/src/lib.rs
```diff
@@ -19,14 +20,44 @@ use syntax::ptr::P;
...
+            .fold(Vec::new(), |mut acc, outs| {
+                acc.extend(outs);
+
+                acc
+            })
+    }
+}
 
 #[plugin_registrar]
 pub fn plugin_registrar(reg: &mut Registry) {
```
llogiq: I'm not sure if I like to split the functionality this way, but I don't have a too strong opinion about it, so this is not something that would keep me from merging.

src/bounded_loop.rs
```diff
@@ -0,0 +1,151 @@
...
+
+            for l in reader.lines() {
+                let line = l.unwrap_or(String::from(""));
+                let splits: Vec<&str> = line.split(",").collect();
+
+                if splits.len() != 2 {
+                    continue;
+                }
+
+                let lid: usize = splits[0].parse().unwrap_or(0usize);
```
llogiq: Is `0` a valid loop ID?

src/bounded_loop.rs
```diff
@@ -0,0 +1,151 @@
...
+
+        File::open("target/mutagen/loops.txt").map(|file| {
+            let reader = BufReader::new(file);
+
+            for l in reader.lines() {
+                let line = l.unwrap_or(String::from(""));
+                let splits: Vec<&str> = line.split(",").collect();
+
+                if splits.len() != 2 {
+                    continue;
```
llogiq: Should we just throw away invalid input? Or perhaps at least warn the user?

src/bounded_loop.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    pub fn id(&self) -> usize {
+        self.0
+    }
+}
+
+impl fmt::Display for LoopId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
```
llogiq: This could simply be `self.0.fmt(f)`, I think.

src/bounded_loop.rs
```diff
@@ -0,0 +1,151 @@
...
+
+impl<'a> LoopStep for LoopCount<'a> {
+    /// step is called on each iteration of a loop
+    fn step(&mut self) {
+        self.count = self.count.saturating_add(1);
+    }
+}
+
+impl<'a> Drop for LoopCount<'a> {
+    fn drop(&mut self) {
```
llogiq: I like how this reads. :+1:

src/bounded_loop.rs
```diff
@@ -0,0 +1,151 @@
...
+        }
+    }
+}
+
+struct BoundCalculator {}
+
+impl BoundCalculator {
+    fn calculate(count: usize) -> usize {
+        let base_value = 1000;  // TODO: Define this values
+        let factor = 1.2;
```
llogiq: A value of 2 is customary here. Perhaps those should both be configurable.

plugin/src/lib.rs
gnieto: I like the fact of having the 2 plugins separated. Even that, I would like to hear for alternatives. Or we can change this on the (near) future.

src/bounded_loop.rs
llogiq: It appears we need yet another file for this. Ah well, let's do this in another PR.

Issue #65, opened 2018-03-20
Mutate while let Some(_) loops


llogiq:
We can mutate `while let Some($ident) = $expr { $block }` loops to extend `$expr` to `if ::mutagen::now($n) { None } else { $expr }`. We already have similar extensions for `if`s, so this should be simple to implement.

rajcspsg:
@llogiq I'm back :-) . Shall I take this issue?
Also, please provide inputs for me to start.

llogiq:
@rajcspsg  It's all yours!

The `fold_expr  function needs another match arm for `ExprKind::WhileLet`. This will give you access to the condition expression, which we subsequently mutate (see the other code for how to do it), the block also needs to be folded using `self.fold_block(..)`.

rajcspsg:
@llogiq _ I started doing changes. When I do changes I tried `cargo build` and I got below error.

```
error[E0277]: the trait bound `usize: std::convert::From<u128>` is not satisfied
   --> src/lib.rs:900:20
    |
900 |             return usize::try_from(val).ok();
    |                    ^^^^^^^^^^^^^^^ the trait `std::convert::From<u128>` is not implemented for `usize`
    |
    = help: the following implementations were found:
              <usize as std::convert::From<u16>>
              <usize as std::convert::From<u8>>
    = note: required because of the requirements on the impl of `std::convert::TryFrom<u128>` for `usize`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.

```

I see rust document has `TryFrom` implemented for `usize` [here](https://doc.rust-lang.org/std/primitive.usize.html#impl-TryFrom%3Ci128%3E).

My rustc version is 

```
$ rustc --version
rustc 1.26.0-nightly (06fa27d7c 2018-04-01)

```

Any idea I get this error?

gnieto:
@rajcspsg I think this was fixed on: bbf53d47cd6b4c897458343d47efcec9dd9d342b

I guess that pulling from master and rebasing your branch should fix the error.

rajcspsg:
@gnieto - Yeah I rebase with upstream and I'm able to do clean build.

llogiq:
After some [research and experimentation](https://llogiq.github.io/2018/05/30/inop.html) I've found this to be impossible to make work in the general case. Sorry!

PR #85, opened 2018-04-13
This closes # 65

rajcspsg:
@llogiq  @gnieto Implement Mutation for [WhileLet Expression](https://github.com/llogiq/mutagen/issues/65).
Sorry for delay of this simple PR.

### DIFF
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -498,6 +498,26 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
             }
             Expr {
                 id,
+                node: ExprKind::WhileLet(pat, cond, block, opt_label),
+                span,
+                attrs,
+            } => {
+                let (n, current, sym, flag, mask) = self.add_mutations(
+                        cond.span,
+                        &["replacing whileLet Some() condition"],
+                    );
+                 
+                let block = if ::mutagen::now($n) { None } else { Some($block) };
+                
+                P(Expr {
+                    id,
+                    node: ExprKind::WhileLet(pat, cond, block, opt_label),
+                    span,
+                    attrs,
+                })
+            }
+            Expr {
+                id,
                 node: ExprKind::ForLoop(pat, expr, block, ident),
                 span,
                 attrs,
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -506,12 +506,14 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
                         cond.span,
                         &["replacing whileLet Some() condition"],
                     );
-                 
-                let block = if ::mutagen::now($n) { None } else { Some($block) };
+               
+                let mut_block = quote_expr!(self.cx(), {
+                    if ::mutagen::now($n) { None } else { Some($block) }
+                });
                 
                 P(Expr {
                     id,
-                    node: ExprKind::WhileLet(pat, cond, block, opt_label),
+                    node: ExprKind::WhileLet(pat, mut_block, block, opt_label),
                     span,
                     attrs,
                 })
--- a/plugin/tests/run-pass/while_let.rs
+++ b/plugin/tests/run-pass/while_let.rs
@@ -6,6 +6,11 @@ extern crate mutagen;
 
 #[mutate]
 fn main() {
+    test_optional();
+}
+
+
+fn test_optional() {
     let mut optional = Some(0);
     while let Some(i) = optional {
         if i > 9 {
@@ -16,6 +21,5 @@ fn main() {
             optional = Some(i + 1);
         }
     };
-    assert!(optional == Some(10));
-
+    assert!(optional == None);
 }
--- a/plugin/tests/run-pass/while_let.rs
+++ b/plugin/tests/run-pass/while_let.rs
@@ -1,0 +1,21 @@
+#![feature(plugin)]
+#![plugin(mutagen_plugin)]
+#![feature(custom_attribute)]
+
+extern crate mutagen;
+
+#[mutate]
+fn main() {
+    let mut optional = Some(0);
+    while let Some(i) = optional {
+        if i > 9 {
+            println!("Greater than 9, quit!");
+            optional = None;
+        } else {
+            println!("`i` is `{:?}`. Try again.", i);
+            optional = Some(i + 1);
+        }
+    };
+    assert!(optional == Some(10));
+
+}


plugin/src/lib.rs
```diff
@@ -496,6 +496,29 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
...
+            } => {
+                let (n, current, sym, flag, mask) = self.add_mutations(
+                        cond.span,
+                        &["replacing while condition with false"],
+                    );
+                let cond = self.fold_expr(cond);
+                let block = fold::noop_fold_block(block, self);
+                let mut_cond = quote_expr!(self.cx(), {
+                    ::mutagen::report_coverage($n..$current, &$sym[$flag], $mask);
+                    ::mutagen::w($cond, $n)
```
gnieto: I think this won't work. Taking this code as example:

```rust
    let mut v = vec![1, 2, 3, 4, 5];
    let mut i = v.iter_mut();

    while let Some(i) = i.next() {
        println!("{}", i);
    }
```

Will be mutated to:

```rust

    let mut v = vec![1, 2, 3, 4, 5];
    let mut i = v.iter_mut();

    while let Some(i) = if { self.now($mutation_count) { false } else { i.next() }  {
        println!("{}", i);
    }
```
This will lead that "if" will have mismatching types: on one branch will return a boolean while the other will return `Option<{integer}>`.

@llogiq If the goal is mutate `while let Some(n) = $e` into `while let Some(n) = None`, it will be the same as no executing the loop, no? Maybe we can just mutate it to `if false { $body }` instead?  I've got the feeling I'm missing something

rajcspsg:
@llogiq @gnieto could you please review my code now.

plugin/src/lib.rs
```diff
@@ -496,6 +496,28 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
...
+                node: ExprKind::WhileLet(pat, cond, block, opt_label),
+                span,
+                attrs,
+            } => {
+                let (n, current, sym, flag, mask) = self.add_mutations(
+                        cond.span,
+                        &["replacing whileLet Some() condition"],
+                    );
+               
+                let mut_block = quote_expr!(self.cx(), {
```
rajcspsg: I have used `quote_expr!`. is it correct?

plugin/tests/run-pass/while_let.rs
```diff
@@ -0,0 +1,25 @@
...
+
+extern crate mutagen;
+
+#[mutate]
+fn main() {
+    test_optional();
+}
+
+
+fn test_optional() {
```
gnieto: You need to add the `#[mutate]` annotation on this function. Otherwise, you are only mutating the `main` function.

plugin/src/lib.rs
gnieto: It's correct

gnieto: You are almost there. Given your example:

```rust
while let Some(i) = optional {
        if i > 9 {
            println!("Greater than 9, quit!");
            optional = None;
        } else {
            println!("`i` is `{:?}`. Try again.", i);
            optional = Some(i + 1);
        }
    };
```

You will get:

```rust
while let Some(i) = if ::mutagen::now(1usize) { None } else {Some(        if i > 9 {
            println!("Greater than 9, quit!");
            optional = None;
        } else {
            println!("`i` is `{:?}`. Try again.", i);
            optional = Some(i + 1);
        }
)} {
        if i > 9 {
            println!("Greater than 9, quit!");
            optional = None;
        } else {
            println!("`i` is `{:?}`. Try again.", i);
            optional = Some(i + 1);
        }
    };
```

I think what you should change this to:

```rust
let mut_expr = quote_expr!(self.cx(), {
    if ::mutagen::now($n) { None } else { $cond }
});
```

Note that you should mutate the expression, which is on the right side. Note also that I've change the variable name on the snippet I propose and you should change it on line 516. With this changes, the example would expand to:


```rust
while let Some(i) = if ::mutagen::now(1usize) { None } else { optional )} {
        if i > 9 {
            println!("Greater than 9, quit!");
            optional = None;
        } else {
            println!("`i` is `{:?}`. Try again.", i);
            optional = Some(i + 1);
        }
    };
```

gnieto: And finally, there's a little last issue. Not all the `while let` expressions are used with `Option`. For example, consider this code:

```rust
fn test_result() {
    let mut optional = Ok(0);
    while let Ok(i) = optional {
        if i > 9 {
            println!("Greater than 9, quit!");
            optional = Err("end");
        } else {
            println!("`i` is `{:?}`. Try again.", i);
            optional = Ok(i + 1);
        }
    };
    assert!(optional == Err("end"));
}
```

Your code will make mutate those expressions and will lead to invalid code, as you will be comparing a `Result` with an `Option` (Look that you are changing the original expression to be `None`, which is of type `Option` and not `Result`).

What you can do is create a function that checks if  `pat` is of type `Some` (we don't have type information, but i think it's good enough, by now, to check if the arm is `Some`). If it does, you mutate the `while let`; otherwise, it should not mutate the expression and return the expression as-is.

Let me know if i'm not clear on some of the points or if you need some help or clarification.

gnieto: I think that something like this would work only for loops like: `while let Some(_) = $cond`, but not on `while let Ok(_) = $cond` or any other pattern.

To be honest, I'm not sure about how we should proceed with this, maybe @llogiq has some idea

llogiq: For the moment I think matching `Some` and `Ok` would be sufficient – and the mutation could use an internal AtomicBool to make the loop body apply once, so that we could have zero or one maximum iterations.

We can make the code generic by creating a trait in `mutagen`. The generated code could look like:

```
while let Some(_) = { static B: AtomicBool = ATOMIC_BOOL_INIT; ::mutagen::wl($expr, $n, &B, ...) } {
    ..
}
```

The `wl` function could then `fetch_or` the `AtomicBool` and generate `None` if the result is `true`. For `Result`s, it needs to `.map_err()` the original Result to `Result<T, ()>` (which is OK, since we throw away the error anyway) and generate `Result::Err(())` to stop the iteration instead.

rajcspsg: @llogiq I've to create new function plugin/lib.rs correct?

llogiq:
I've now looked into your code – you seem not to have pushed your latest changes. We will also need to make this mutation opportunistic (because someone could have their own `Some(_)` variant outside of `Option<_>`, same goes for `Error`. I'm going to add a function to the helper library that we can use.

llogiq:
After some [research and experimentation](https://llogiq.github.io/2018/05/30/inop.html) I've come to the conclusion that this is either impossible or a special case of statement reduction. I'm afraid the way I asked you to implement this will lead to errors on some code; this is something we won't allow in mutagen, because if the build breaks we can no longer test.

PR #86, opened 2018-04-14
mutate ref muts by cloning

llogiq:
This is one opportunistic mutation that was missing so far. It also moves coverage reporting into the trait impls so we only get reports for effective opportunistic mutations.

However, I run into the same problem as elsewhere when trying to report coverage in `fold_first_block`. Strange.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -66,7 +66,7 @@ fn write_binop_arm(out: &mut Write,
     if shift {
         writeln!(out, "
             BinOpKind::{0} => {{
-                let (n, current, sym, flag, mask) = p.add_mutations(
+                let (n, _current, sym, flag, mask) = p.add_mutations(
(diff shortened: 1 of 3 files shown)


gnieto:
I think I found the bug. We have 2 `add_mutations` functions, one on `MutationPlugin`, which increases the coverage_counter; the other is located on `Mutator`, which does not increase the `coverage_counter`.
On `fold_first_block` we are using the `Mutator` one instead of `MutationPlugin`. Then, when setting the size for the coverage symbol, we detect that there are no mutations and the symbol is replaced by `()`.

llogiq:
Great detective work, @gnieto – I think I'll be able to fix this shortly.

llogiq:
This is now complete and passes tests.

PR #87, opened 2018-04-16
Fix mutation for interchangeables.

adeschamps:
Previously this was producing mutations like this, where the new variables go out of scope immediately and therefore have no effect.

```rust
if ::mutagen::now($n) {
    let (a, b) = (b, a);
}
```

Now it produces this, which I think is a nice example of variable shadowing being useful:
```rust
let (a, b) = if ::mutagen::now($n) {
    (b, a)
} else {
    (a, b)
};
```


llogiq:
Good catch! :+1:

PR #95, opened 2018-05-01
Failing test for MayClone

gnieto:
A method of a struct which recives `&mut self` gets an error when
mutated.

This is because the `MayClone` mutation as the code generated is:

```rust
            let self = if ::mutagen::MayClone::may_clone(self) {
                _self_clone =
                    ::mutagen::MayClone::clone(self, 2usize,
&__COVERAGE1[0usize], 2usize);
                &mut _self_clone
            } else {
                self
            };
```

This leads to initialize a variable with the name "self", which throws
this error: "expected unit struct/variant or constant, found local
variable `self` [E0424]"


llogiq:
Thank you. Fix incoming shortly.

PR #97, opened 2018-05-03
extend to cope with other non-maximal consts

### LINKED ISSUE #93: Extend `int_constant_can_add_one` to cope with other non-maximal consts
In [`plugin/src/lib.rs`, line 621](https://github.com/llogiq/mutagen/blob/dffca47b7957cfce47d136436a21dde63ed2dbf6/plugin/src/lib.rs#L621), we only check against u8::MAX, however this doesn't tell us the whole truth. If our number is indeed *greater* than u8::MAX, we can assume that the type is at least 16 bit, so we should return `true` unless the given value is equal to *any* MAX value.

This can be done most easily by creating a static array of MAX values and using `slice::contains` to check for any value. I'm up for mentoring if anyone wants and/or needs it.

rleungx:
Closes #93.

### DIFF
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -615,9 +615,21 @@ fn int_constant_can_subtract_one(i: i64, ty: LitIntType) -> bool {
     i as i64 > min
 }
 
+static MAX_VALUES: &[u64] = &[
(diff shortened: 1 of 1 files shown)


llogiq:
Thank you!

PR #105, opened 2018-05-05
support 128 bit types

### LINKED ISSUE #98: Change number handling to use 128-bit types
If I recall correctly, `u128`/`i128` were recently stabilized, so our `int_constant_can_add_one`/`int_constant_can_subtract_one` functions (in `plugin/src/lib.rs`) should be updated to allow for those types.

A good test case would be a small program (in `plugins/tests/run-pass`) that contains the constants `0u128`, `std::u128::MAX`, `42u128` (and likewise `std::i128::MIN`, `0i128`, `std::i128::MAX`), but as plain numbers. The program does not need to do anything useful.

rleungx:
Closes #98.


llogiq:
Thank you!

PR #107, opened 2018-05-14
Make self / Self interchange work

### LINKED ISSUE #103: Extend interchange to work with self types
Currently, the pattern destructuring does not take `self` vs. `Self` into account. We could go even further on `impl`s (if we see them) and store their type, so that a method `Foo::fu(self, other: Foo)` can exchange arguments.

llogiq:
This fixes #103 

This makes both `Hash` and `PartialEq` implementations treat `self` and `_: Self` as equal, including putting them in the lifetime whitelist, because even if `Self` contains a lifetime, it's generally fixed and thus the same for all arguments.

While testing, I also noticed that putting `#[mutate]` on `impl` blocks didn't work, so I fixed that, too, while I was at it.

### DIFF
--- a/plugin/src/lib.rs
+++ b/plugin/src/lib.rs
@@ -404,7 +404,7 @@ impl<'a, 'cx> Folder for MutatorPlugin<'a, 'cx> {
                     generics,
                     opt_trait_ref,
                     ty,
-                    impl_items,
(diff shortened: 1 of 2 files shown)


llogiq:
(Also it just occurred to me that I passed up the chance to name the branch "exchange-*thy*self". Sad!)

PR #108, opened 2018-05-14
Apply some clippy suggestions

gnieto:
Apply some clippy suggestions.

I'm not sure about the changes on: https://github.com/llogiq/mutagen/commit/fa40e2837b9233dc38da91535ea5c1a29456acfe  (I've got the feeling that it may cause issues with the borrow checker. Even that, I didn't found any issue with any mutated code).

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -239,7 +239,7 @@ pub fn fold_binop(p: &mut MutatorPlugin, id: NodeId, op: BinOp, left: P<Expr>, r
                 );
             quote_expr!(p.cx(), {{
                 ::mutagen::report_coverage($n..$current, &$sym[$flag], $mask);
-                ::mutagen::eq($left, $right, $n)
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -289,7 +289,7 @@ pub fn fold_binop(p: &mut MutatorPlugin, id: NodeId, op: BinOp, left: P<Expr>, r
                 );
             quote_expr!(p.cx(), {{
                 ::mutagen::report_coverage($n..$current, &$sym[$flag], $mask);
-                ::mutagen::gt($right, $left, $n)
+                ::mutagen::gt(&$right, &$left, $n)
```
llogiq: Which lint does this appease?

llogiq: Ah, `needless_pass_by_value`. OK then.

gnieto: Sorry, I missed the notification about the comment above.

PR #109, opened 2018-05-21
Avoid redundant mutations

### LINKED ISSUE #41: Detect and Avoid Equal Mutations
Unlike most mutation testing, we want to go fast. Thus avoiding work should be
one goal to pursue. Thus, detecting and avoiding equal mutations would be a
win if the detection takes less time that running the test suite (which rules
out complex analysis, but that isn't needed for some quick wins).

For example, we currently mutate if-conditions with `true`, `false` and `!cond`
– however, we will also mutate any sub-conditions, so if we have
`if x || y { .. }`, we will generate the following mutations:

1. replace if condition `x || y` by true
2. replace if condition `x || y` by false
3. invert if condition `x || y` (resulting in `!(x || y)`)
4. replace `x || y` with true (equal to 1)
5. replace `x || y` with false (equal to 2)
6. replace `x || y` with `x`
7. replace `x || y` with `!x`
8. replace `x || y` with `x || !y`

So we could remove 2 of those 8 mutations without any ill effects. Worse, if
`x` is another expression that we mutate (for example `a == b`), then its
mutations may also be equal to the one above, because `true || x` == `true`. We
should perhaps consider a more careful approach in adding mutations, so that
already issued mutations are avoided.

### LINKED ISSUE #53: Add mutation codes
Currently the mutation descriptions are mostly nice to read, but quite wordy and not very machine-understandable. In order to e.g. allow filtering mutations, we should add a unique mutation code per mutation type.

This is fairly easy: Just prepend the code before each message in `add_mutations` in the plugin. The hard part is, as usual, naming things.

gnieto:
Opening this to get (late) feedback. This mainly contains three parts (I can split each commit into distinct PRs):

# Add mutation ID to `mutations.txt`
Added the mutation ID and parse it on runner, in order to be able to skip some mutations. This is needed as we previously infered the mutation count from the line count on `mutations.txt`.

# Add mutation types

This should close #53 . I've added this on the scope of this PR as it will enable to do not emit certain mutation types depending on some situations

# Avoid redundant mutations

This should close #41 .The system will have a stack of mutation types to avoid. On each call to `fold_expr`, we will add an entry to this stack and, when leaving, we will pop it. When we
are performing a mutation, we are able to impose some restrictions to
the direct sub expressions.

So, if we are mutating an `&&`, we can force to left and
right expressions to do not emit a `replace with false` mutation, as we
will already mutate this and to `false`, and that would lead to a
redundant mutation.

:warning: I'm aware build is failing, I will fix tests if this gets approved.


llogiq:
This is looking very good :+1:. In fact I got to read a paper about avoiding duplicate mutations that suggests a very similar approach. Approved modulo CI failures.

gnieto:
CI fixed.

PR #121, opened 2018-06-28
(WIP) refactor: Move reporting into mutagen calls

llogiq:
This also uncovered a bug in the "exchange arguments" mutation, which is also fixed. However, there may now be some interference between that and the "clone mut ref arg" mutation, which I intend to tackle next.


src/lib.rs
```diff
@@ -175,38 +176,45 @@ pub fn next() {
...
-pub fn gt<R, T: PartialOrd<R>>(x: &T, y: &R, n: usize) -> bool {
+pub fn gt<R, T: PartialOrd<R>>(x: &T, y: &R, n: usize, flag: &AtomicUsize, mask: usize) -> bool {
+    report_coverage(n..(n + 7), flag, mask);
     MU.gt(x, y, n)
 }
 
 /// use instead of `>=` (or, switching operand order `<=`)
-pub fn ge<R, T: PartialOrd<R>>(x: &T, y: &R, n: usize) -> bool {
+pub fn ge<R, T: PartialOrd<R>>(x: &T, y: &R, n: usize, flag: &AtomicUsize, mask: usize) -> bool {
+    report_coverage(n..(n + 7), flag, mask);
```
gnieto: I think this constant offsets may be to fragile. For example, if at some point the amount of mutations for comparisons changes, we may emit an incorrect coverage. What do you think about pass this offsets by parameter (the callers usually have the list of mutations, no?)?

src/lib.rs
```diff
@@ -160,12 +159,14 @@ pub fn get() -> usize {
 /// check if the argument matches the current mutation count
 ///
 /// this simplifies operations like `if mutagen::MU.now(42) { .. }`
-pub fn now(n: usize) -> bool {
+pub fn now(n: usize, flag: &AtomicUsize, mask: usize) -> bool {
+    report_coverage(n..(n + 1), flag, mask);
```
gnieto: What about pushing the report to `MU.now(...)` and `MU.diff(...)`? This way we will make sure that future implementors will report the coverage.

llogiq: This would embed the coverage reporting within the mutation code, which may hinder testability. Otherwise I have no preference either way, but this was the easiest way to refactor it.

PR #123, opened 2018-06-29
fix interference between arg interchange & ref mut clone

### LINKED ISSUE #122: Interference between Exchange Args and Mut ref args clone mutations
If a `self` arg will be both exchanged and cloned (in different mutations) the latter mutation will make the former ineffective.

Solving this requires tracking if `self' was exchanged and changing the `ident` accordingly.

llogiq:
This fixes #122


llogiq:
It appears we need yet another rustup. `Symbol`s are now `Ident`s with `Span`.

llogiq:
Rebased. I want this merged before publishing another version to fix #124

gnieto:
Maybe it would be nice to add a little test with the failing example to avoid regressions.

Issue #127, opened 2018-07-04
error[E0277]: the size for value values of type `[u8]` cannot be known at compilation time


bbigras:
mutagen dfc844f597db911f4231d09c266a1462a9240cfd
nightly-x86_64-unknown-linux-gnu (default)
rustc 1.28.0-nightly (e3bf634e0 2018-06-28)

Here's a simplified case.

```rust
#[cfg_attr(test, mutate)]
fn something() -> bool {
    let n = "test".as_bytes();
    *n == *b"test"
}
```

```
error[E0277]: the size for value values of type `[u8]` cannot be known at compilation time
   --> src/main.rs:784:18
    |
784 | #[cfg_attr(test, mutate)]
    |                  ^^^^^^^ doesn't have a size known at compile-time
    |
    = help: the trait `std::marker::Sized` is not implemented for `[u8]`
    = note: to learn more, visit <https://doc.rust-lang.org/book/second-edition/ch19-04-advanced-types.html#dynamically-sized-types--sized>
    = note: required by `mutagen::eq`
```

PR #129, opened 2018-07-06
Fix issue with references and type-inference

gnieto:
Tentatively fixes https://github.com/llogiq/mutagen/issues/127

When we call to `::mutagen::eq`, we are borrowing the left and right
  values (so, we are `::mutagen::eq(&x, &y, ...)`. The previous
signature was also forcing a borrow `::mutagen::eq(x: &T, y: &R)`, which
provoked some type-inference issues as reported on:
https://github.com/llogiq/mutagen/issues/127.

### DIFF
--- a/plugin/tests/run-pass/issue-127.rs
+++ b/plugin/tests/run-pass/issue-127.rs
@@ -1,0 +1,25 @@
+#![feature(plugin)]
+#![plugin(mutagen_plugin)]
+#![feature(custom_attribute)]
+
+extern crate mutagen;
+
+#[mutate]
+fn something() -> bool {
+    let n = "test".as_bytes();
+    *n == *b"test"
+}
+
+#[mutate]
+fn owned_eq() -> Option<String> {
+    let x = "Hello".to_string();
+    if x == "Hello" {
+        Some(x)
+    } else {
+        None
+    }
+}
+
+fn main() {
+    something();
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,7 +93,7 @@ impl Mutagen {
     }
 
     /// use instead of `==`
-    pub fn eq<R, T: PartialEq<R>>(&self, x: &T, y: &R, n: usize) -> bool {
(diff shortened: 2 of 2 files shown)


llogiq:
How about adding a test as the following:

```
struct Me;
struct You;

impl PartialEq<You> for Me {
   fn partial_eq(&self, &You) -> Option<bool> { Some(true); } // I am YOU! :-D
}

#[mutate]
fn are_we_equal(me: Me, you: You) {
    let _ = me == you;
}
```

gnieto:
Wow, I feel embarased: I knew that and I somehow missed it... Sorry for that!

Anyway, I think that with the current change, I'm fixing it, which is mainly pass the generic type without reference. What do you think now?

llogiq:
I may have another illuminating test case:

```rust
#[mutate]
fn owned_eq() -> Option<String> {
    let x = "Hello".to_string();
    if x == "Hello" {
        Some(x)
    } else {
         None
    }
}
```

gnieto:
The given test case seems to compile with the proposed version. Here's the expanded version:

```rust
fn owned_eq() -> Option<String> {
    static __LOOP_COUNTERS: [::std::sync::atomic::AtomicUsize; 0usize] = [];
    static __COVERAGE5: [::std::sync::atomic::AtomicUsize; 1usize] =
        [::std::sync::atomic::ATOMIC_USIZE_INIT];
    if let Some(r) =
           ::mutagen::Defaulter::get_default(5usize, &__COVERAGE5[0usize],
                                             1usize) {
        return r;
    }
    let x = "Hello".to_string();
    if ::mutagen::t(::mutagen::eq(&x, &"Hello", 6usize, &__COVERAGE5[0usize],
                                  2usize), 9usize, &__COVERAGE5[0usize],
                    4usize) {
        Some(x)
    } else { None }
}
```

Anyway, if you want, I can open a PR with the examples and add them to the test suite and close this PR.

gnieto:
Rebased and added the test case

PR #130, opened 2018-07-09
Fix rust update breakage

gnieto:
A new `NodeId` was added for impl traits. I don't know if it makes sense to check equality or hashing for node ids.


llogiq:
No, we can safely ignore `NodeId`s for hashing and equality. This looks good to me.

PR #138, opened 2019-01-26
WIP: Rewrite plugin to proc_macro_attribute, ...

llogiq:
replace build.rs scripts with macros (which should improve our build time), change some internal logic.

I may have introduced some bugs, and need to finish some todos, but I'm quite happy with most of the changes.


power-fungus:
I have an issue with building this branch. `cargo build` fails with

```
error[E0599]: no method named `unwrap` found for type `proc_macro2::Span` in the current scope
   --> src/lib.rs:187:33
    |
187 |         let nightly_span = span.unwrap();
    |                                 ^^^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0599`.
error: Could not compile `mutagen-plugin`.
```

What have I done wrong?

power-fungus:
turns out: this issue (and similar ones) is caused by a too old version of `proc-macro2`. For me, everything before `proc-macro-0.4.27` results in failures as certain functions/names could not be found.

llogiq:
Yes, there are still a few errors. I just pushed a new version with some changes before I saw your PRs, so sorry if they are already stale.

Issue #142, opened 2019-02-18
Mutagen architecure proposal


power-fungus:
Hi, I've been working on re-structuring your project `mutagen` to be more production ready. 
https://github.com/power-fungus/mutagen-preview

I spent the last weeks reading into the current project and re-writing parts of mutagen using the same interface, but with a different architecture. It's not completed yet but should give a rough outline of what the finished project will look like: the backbone and glue code already exists and will probably not change too widely.

@llogiq **I would like to know if you are interested in my approach and if you would consider merging my proposal in your project**. If so, I will finish this preview into production quality in the next months. Also I am open to any discussion about interface and design decisions.

I would like to have this rewrite published (once finished) under the "brand" of `mutagen`:
* The interface is the same
  * mutators are introduced using the attribute `#[mutate]`
  * mutations are selected at run-time using the environment variable `MUTATION_ID` (renamed for clarity)
  * The target code is compiled only once
* the rewrite is true to the original goals of `mutagen`
* the used technologies is the same to the code in the `proc-macro-attribute` branch
  * new procedural macro attribute
  * uses `syn` and `quote`

For details about the architecture, see: https://github.com/power-fungus/mutagen-preview/tree/master/docs
Main Benefits are:
* Debugging of the code transformation will be easier due to modular transformation
* The behavior of the mutations can be tested a standard test suite by faking the global `MUTATION_ID` variable.
* it will be possible to add customization via arguments to the `#[mutate]` attribute
* community collaboration will be easier since mutators and transformers are small bits of code.
* mutators can be documented independent from each other.

@llogiq I would like to read from you soon! :)

llogiq:
The code can be tested using what?

I'm a bit short on time right now, but I'll review your code during the next week, I promise!

power-fungus:
Next week is fine  :+1: 
I updated the post above to complete the previously incomplete sentence about testing

power-fungus:
misclick :(

llogiq:
So if I understand your proposal right, you want to decouple various mutation-embedding parts of the plugin so that each implements its own `Fold`.

What I like:

* The code looks pretty simple
* The transformation abstraction affords us a structured way to switch transformations on or off (on the other hand, we also could extend our current restrictions system for that)
* Future transformations should be easy enough to extend

What I dislike:

[...]

PR #149, opened 2019-07-16
proposal and preview for mutagen v0.2

power-fungus:
As discussed in #142, this is a proposal for a modular architecture, which is the base for `mutagen v0.2`. I reviewed all parts of the code and tried to move code that works fine to the new system.

This is a minimal implementation to demonstrate the effectiveness of the architecture. The steps towards releasing `mutagen v0.2` are:

- [ ]: implement more mutators and document their functionality
- [ ]: generate meaningful reports

Are there any other major tasks for this PR?


power-fungus:
I have done quite a few changes to the architecture since the first commit. I am quite happy with the current state.

* I added a crate `mutagen-core` that contains most code. This crate is imported from `mutagen-transform` and `mutagen`. This makes re-use of code a lot easier and still meets the requirement of a standalone crate for the procedural macro.
* The code to transform expressions is defined in the same file with their mutators.
* The hack for exhaustive testing has beed replaced by the module `mutagen-selftest`, which provides the same functionality in a cleaner way.
* I added an example-crate with well-known functions that help demonstrate the use of `#[mutate]` and the functionality of the runner. The runner can be called with `cargo run --bin cargo-mutagen --package cargo-mutagen` from the `examples/simple` directory

README.md
```diff
@@ -1,73 +1,68 @@
...
-
-### How mutagen Works
-
-Mutagen works as a procedural macro. This means two things:
-
-1. You'll need a nightly rust toolchain to compile the plugin.
-2. it only gets to see the code you mark up with the `#[mutate]` annotation, nothing more.
+A change (mutation) in the program source code is most likely a bug of some kind. A good test suite can detect changes in the source code by failing ("killing" the mutant). If the test suite is green even if the program is mutated (the mutant survives), the tests fail to detect this bug.
 
-It also will only see the bare AST, no inferred types, no control flow or data flow, unless we analyse them ourselves. But not only that, we want to be *fast*.  This means we want to avoid doing one compile run per mutation, so we try to bake in all mutations into the code once and select them at runtime via a mutation count. This means we must avoid mutations that break the code so it no longer compiles.
```
llogiq: Why are we losing this information? I feel it is a crucial piece to understand the design decisions for mutagen, which haven't really changed with your redesign.

README.md
```diff
@@ -1,73 +1,68 @@
...
+*Use `#[mutate]` for tests only.* This is done by always annotating functions or modules with `#[cfg_attr(test, mutate)]` instead, which applies the `#[mutate]` annotation only in `test` mode. If a function is annotated with plain `#[mutate]` in every mode, the mutation-code is baked into the code even when compiled for release versions. However, when using `mutagen` as `dev-dependency`, adding a plain `#[mutate]` attribute will result in compilation errors in non-test mode since the compiler does not find the annotation.
 
-You can run `cargo mutagen -- --coverage` in order to reduce the time it takes to run the mutated code. When running on this mode, it runs the testsuite at the beginning of the process and checks which tests are hitting mutated code. Then, for each mutation, instead of running the whole testsuite again, it executes only the tests that are affected by the current mutation. This mode is specially useful when the testsuite is slow or when the mutated code affects a little part of it.
+*Use `mutagen` as `dev-dependency`, unless otherwise necessary.* Compiling `mutagen` is time-intensive and library-users should not have to download `mutagen` as a dependency.
```
llogiq: Perhaps we should advise users to use `sccache`?

mutagen-core/src/mutator/mutator_binop_add.rs
```diff
@@ -0,0 +1,95 @@
...
+pub struct MutatorBinopAdd {}
+
+impl MutatorBinopAdd {
+    pub fn run<L: Add<R>, R>(
+        mutator_id: u32,
+        left: L,
+        right: R,
+        runtime: MutagenRuntimeConfig,
+    ) -> <L as Add<R>>::Output {
+        if runtime.mutation_id != mutator_id {
```
llogiq: Do I understand right that a Mutator can carry out only a single mutation? The current design allows a single mutation to add several code paths.

README.md
power-fungus: I have no experience with that. I wanted to say that if a library uses `mutagen`, users of that library should not have to compile mutagen since it is only used for internal tests. 

For testing purposes, the compile-time is fine. I guess I have to rewrite this section to avoid confusion about this topic.

mutagen-core/src/mutator/mutator_binop_add.rs
power-fungus: It was always the idea to allow multiple mutations per mutator and I implemented this for the mutator `lit_int` in the newest commit. This required more changes than expected.

README.md
power-fungus: This information is indeed important. I thought I kept this information somewhere. I will add this paragraph back into the documentation and the readme.

docs/mutators/lit_int.md
```diff
@@ -6,11 +6,12 @@ Integer literals like `0`, `1u8`, `5isize`
 
 ## Mutations
 
-1. replacing the literal by with a different literal
+1. replacing the literal with a different literal
```
llogiq: I think we should be able to make integer literal mutation optimistic. This would also allow us to support 128-bit types.

mutagen-core/src/mutator/mutator_binop_add.rs
llogiq: :+1:

docs/mutators/lit_int.md
power-fungus: I agree. The issue is that the method to get the value of the literal `value()` returns a `u64`. (https://docs.rs/syn/0.15.40/syn/struct.LitInt.html). 128-bit literals are trivially supported if their values fit inside a `u64`.
I am currently investigating what happens within `syn` in case of overflows.

power-fungus: I agree. The issue is that the method to get the value of the literal `value()` returns a `u64`. (https://docs.rs/syn/0.15.40/syn/struct.LitInt.html). 128-bit literals are trivially supported if their values fit inside a `u64`.
I am currently investigating what happens within `syn` in case of overflows.

README.md
power-fungus: Updated in the newest commits

README.md
power-fungus: In the newest commits this section is back in the readme. The documents about the architecture and design decisions also contain paragraphs on this topic.

docs/mutators/lit_int.md
llogiq: We can get the Token::Lit instead & get the String value from there.

power-fungus: During my research, I stumbled apon [your PR](https://github.com/dtolnay/syn/pull/574) towards this issue. I agree with the author of syn on the plans for the new api.
I suggest deferring this issue until `syn-1.0` has landed. As you know, this affects only literals with a value larger than `u64::max_value()` and not all values of type `u128`. This behavior is documented in [`syn::LitInt`](https://docs.rs/syn/0.15.42/syn/struct.LitInt.html).

llogiq: :+1:

power-fungus:
In the newest commit, I implemented another mutator that mutates comparison operators or `PartialOrd` into each other. I have some of other mutators planned, but I want to look at the current structure a bit and try to find rough edges in the design before implementing more mutators.

At this stage, it is still possible to change the architecture of everything. Do you have any requests / suggestions?

.travis.yml
```diff
@@ -6,13 +6,14 @@ os:
...
+  - cargo test -p mutagen-core # run unit tests first
+  - cargo test -p mutagen-selftest # run integration tests
+  - cargo test # run all (other) tests
+
+  # run executable cargo-mutagen on test project
+  - pushd examples/simple
+  - cargo run --bin cargo-mutagen --package cargo-mutagen
+
+  # check format
+  cargo fmt -- --check
```
power-fungus: Is this a good idea? I've seen this in a blogpost. However, I also seen arguments against this.

llogiq: I'm not opposed to it. Let's leave it in and see if it burdens us too much.

mutagen-core/src/lib.rs
```diff
@@ -0,0 +1,16 @@
+#![feature(proc_macro_span)]
+#![feature(box_syntax)]
```
llogiq: Where is this needed?

mutagen-core/src/lib.rs
```diff
@@ -0,0 +1,16 @@
...
+#![feature(vec_remove_item)]
```
llogiq: Also this?

mutagen-core/src/optimistic/add_to_sub.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+impl<L, R> AddToSub<R> for L
+where
+    L: Add<R>,
+{
+    type Output = <L as Add<R>>::Output;
+
+    default fn may_sub(self, _r: R) -> <L as Add<R>>::Output {
+        panic!("not sub");
```
llogiq: Why do we panic here?

llogiq:
Sorry for letting you wait for so long. I was on vacation. This looks pretty good so far, I only commented on a few questions I had.

I think this should be ready to merge once we reach feature parity with the previous version. Or would you like this to be merged sooner?

mutagen-core/src/lib.rs
power-fungus: This is not strictly needed. This is a leftover from converting to boxed closures and getting those to work. I can revert this if necessary.

mutagen-core/src/optimistic/add_to_sub.rs
power-fungus: As far as I understand:

* If the optimistic assumption fails, the mutant should not be considered survived.
* To be able to kill the mutant, the test suite has to fail.
* To fail the test suite, the code must have some different behavior when activating the mutant. 
* I believe panicking is the only/easiest way to cause behavior that is different from the non-mutated case. However, I am open to other options.

In the future, I would like to read the output of the test-binary and detect if a certain panic-message has been printed. Based on that, the mutant could be considered "failed" instead of "killed" or "survived".

mutagen-core/src/lib.rs
llogiq: Would be nice, otherwise we can remove it in a later PR. I usually try to minimize use of nightly features, that's why I asked.

mutagen-core/src/lib.rs
power-fungus: This is used during configuration of the transformers. I could replace its use with the 2-line implementation of `remove_item`. We are on nightly already and I think those features are quite stable.

The features `proc_macro_span` and `specialization` are necessary for the current approach to be implementable. I thought that the extra features do not add extra risk to the project.

mutagen-core/src/optimistic/add_to_sub.rs
llogiq: Ok, that makes sense. I had envisioned a backchannel (an mmap, socket or simple file) to inform the test runner of the ineffective mutation on the no-mutation-run, so the runner could mark the mutation as ineffective. However, there is actually a tri-state at play here, because such a mutation could be within a generic function that monomorphizes to any number of ineffective mutations. Thus, we should be able to both mark an optimistic mutation as (potentially) ineffective and as effective (at least we should do so if it was marked as ineffective before and succeeds nonetheless).

llogiq: I should add that in-band messaging is inherently problematic and I would oppose misusing panics for that purpose. What if a user has a custom panic message that is inadvertently parsed as a killed mutant?

mutagen-core/src/lib.rs
llogiq: Agreed.

mutagen-core/src/lib.rs
power-fungus: feature removed.

mutagen-core/src/optimistic/add_to_sub.rs
power-fungus: I see that using panics might not be the cleanest solution. I will work on some backchannel for coverage-analysis. The information about optimistic assumptions can be included there.

However, I think that not necessarily part of a first working release, which I would like to finish in early September. 

I did not think about optimistic mutantions in generic functions. I will investigate some issues here.

llogiq: I'm fine with that. Currently we just write a file on the nonmutated run, but there are likely faster and leaner solutions.

power-fungus:
I spend probably more time than I should have on getting a good progress bar inspired by the one cargo uses. I got it working so far.

running `cargo run -p cargo-mutagen` in the `examples/simple` directory gives the following report on my machine: 

<details><summary>stdout</summary>
<p>

```
1: lit_int, replace `1` with `2`, at examples/simple/src/add1.rs@5:4-5:5(fn simple_add) ... killed
2: lit_int, replace `1` with `0`, at examples/simple/src/add1.rs@5:4-5:5(fn simple_add) ... killed
3: lit_int, replace `2` with `3`, at examples/simple/src/add1.rs@5:8-5:9(fn simple_add) ... killed
4: lit_int, replace `2` with `1`, at examples/simple/src/add1.rs@5:8-5:9(fn simple_add) ... killed
5: binop_add, replace `+` with `-`, at examples/simple/src/add1.rs@5:6-5:7(fn simple_add) ... killed
6: lit_int, replace `1` with `2`, at examples/simple/src/add1.rs@10:4-10:7(fn simple_add_u8) ... killed
7: lit_int, replace `1` with `0`, at examples/simple/src/add1.rs@10:4-10:7(fn simple_add_u8) ... killed
8: lit_int, replace `2` with `3`, at examples/simple/src/add1.rs@10:10-10:11(fn simple_add_u8) ... killed
9: lit_int, replace `2` with `1`, at examples/simple/src/add1.rs@10:10-10:11(fn simple_add_u8) ... killed
10: binop_add, replace `+` with `-`, at examples/simple/src/add1.rs@10:8-10:9(fn simple_add_u8) ... killed
11: lit_int, replace `1` with `2`, at examples/simple/src/bubblesort.rs@6:13-6:14(fn bubblesort_for) ... killed
12: lit_int, replace `1` with `0`, at examples/simple/src/bubblesort.rs@6:13-6:14(fn bubblesort_for) ... SURVIVED
13: lit_int, replace `1` with `2`, at examples/simple/src/bubblesort.rs@7:17-7:18(fn bubblesort_for) ... killed
14: lit_int, replace `1` with `0`, at examples/simple/src/bubblesort.rs@7:17-7:18(fn bubblesort_for) ... killed
15: lit_int, replace `1` with `2`, at examples/simple/src/bubblesort.rs@8:23-8:24(fn bubblesort_for) ... killed
16: lit_int, replace `1` with `0`, at examples/simple/src/bubblesort.rs@8:23-8:24(fn bubblesort_for) ... killed
17: binop_cmp, replace `>` with `<`, at examples/simple/src/bubblesort.rs@8:26-8:27(fn bubblesort_for) ... killed
18: binop_cmp, replace `>` with `<=`, at examples/simple/src/bubblesort.rs@8:26-8:27(fn bubblesort_for) ... killed
19: binop_cmp, replace `>` with `>=`, at examples/simple/src/bubblesort.rs@8:26-8:27(fn bubblesort_for) ... SURVIVED
20: lit_int, replace `1` with `2`, at examples/simple/src/bubblesort.rs@9:29-9:30(fn bubblesort_for) ... killed
21: lit_int, replace `1` with `0`, at examples/simple/src/bubblesort.rs@9:29-9:30(fn bubblesort_for) ... killed
22: lit_bool, replace `true` with `false`, at examples/simple/src/bubblesort.rs@18:21-18:25(fn bubblesort_while) ... killed
23: lit_bool, replace `false` with `true`, at examples/simple/src/bubblesort.rs@20:17-20:22(fn bubblesort_while) ... killed (timeout)
24: lit_int, replace `1` with `2`, at examples/simple/src/bubblesort.rs@21:17-21:18(fn bubblesort_while) ... killed
25: lit_int, replace `1` with `0`, at examples/simple/src/bubblesort.rs@21:17-21:18(fn bubblesort_while) ... killed
26: lit_int, replace `1` with `2`, at examples/simple/src/bubblesort.rs@22:23-22:24(fn bubblesort_while) ... killed
27: lit_int, replace `1` with `0`, at examples/simple/src/bubblesort.rs@22:23-22:24(fn bubblesort_while) ... killed
28: binop_cmp, replace `>` with `<`, at examples/simple/src/bubblesort.rs@22:26-22:27(fn bubblesort_while) ... killed
29: binop_cmp, replace `>` with `<=`, at examples/simple/src/bubblesort.rs@22:26-22:27(fn bubblesort_while) ... killed
30: binop_cmp, replace `>` with `>=`, at examples/simple/src/bubblesort.rs@22:26-22:27(fn bubblesort_while) ... SURVIVED
31: lit_int, replace `1` with `2`, at examples/simple/src/bubblesort.rs@23:29-23:30(fn bubblesort_while) ... killed
32: lit_int, replace `1` with `0`, at examples/simple/src/bubblesort.rs@23:29-23:30(fn bubblesort_while) ... killed (timeout)
33: lit_bool, replace `true` with `false`, at examples/simple/src/bubblesort.rs@24:25-24:29(fn bubblesort_while) ... killed
34: lit_int, replace `0` with `1`, at examples/simple/src/ggt.rs@6:16-6:17(fn ggt_loop) ... killed (timeout)
35: binop_eq, replace `==` with `!=`, at examples/simple/src/ggt.rs@6:13-6:15(fn ggt_loop) ... killed (timeout)
36: lit_int, replace `0` with `1`, at examples/simple/src/ggt.rs@9:16-9:17(fn ggt_loop) ... killed (timeout)
37: binop_eq, replace `==` with `!=`, at examples/simple/src/ggt.rs@9:13-9:15(fn ggt_loop) ... killed (timeout)
38: binop_cmp, replace `>` with `<`, at examples/simple/src/ggt.rs@12:13-12:14(fn ggt_loop) ... killed
39: binop_cmp, replace `>` with `<=`, at examples/simple/src/ggt.rs@12:13-12:14(fn ggt_loop) ... killed
40: binop_cmp, replace `>` with `>=`, at examples/simple/src/ggt.rs@12:13-12:14(fn ggt_loop) ... SURVIVED
41: binop_eq, replace `==` with `!=`, at examples/simple/src/ggt.rs@22:9-22:11(fn ggt_rec) ... killed
42: lit_int, replace `0` with `1`, at examples/simple/src/ggt.rs@22:22-22:23(fn ggt_rec) ... killed
43: binop_eq, replace `==` with `!=`, at examples/simple/src/ggt.rs@22:19-22:21(fn ggt_rec) ... killed
44: binop_bool, replace `||` with `&&`, at examples/simple/src/ggt.rs@22:14-22:16(fn ggt_rec) ... SURVIVED
45: lit_int, replace `0` with `1`, at examples/simple/src/ggt.rs@22:32-22:33(fn ggt_rec) ... killed
46: binop_eq, replace `==` with `!=`, at examples/simple/src/ggt.rs@22:29-22:31(fn ggt_rec) ... killed
47: binop_bool, replace `||` with `&&`, at examples/simple/src/ggt.rs@22:24-22:26(fn ggt_rec) ... killed
48: binop_cmp, replace `>` with `<`, at examples/simple/src/ggt.rs@25:9-25:10(fn ggt_rec) ... killed
49: binop_cmp, replace `>` with `<=`, at examples/simple/src/ggt.rs@25:9-25:10(fn ggt_rec) ... killed
50: binop_cmp, replace `>` with `>=`, at examples/simple/src/ggt.rs@25:9-25:10(fn ggt_rec) ... SURVIVED
51: lit_int, replace `1` with `2`, at examples/simple/src/not_covered.rs@5:4-5:5(fn simple_assert_not_covered) ... NOT COVERED
52: lit_int, replace `1` with `0`, at examples/simple/src/not_covered.rs@5:4-5:5(fn simple_assert_not_covered) ... NOT COVERED
53: lit_int, replace `3` with `4`, at examples/simple/src/not_covered.rs@5:8-5:9(fn simple_assert_not_covered) ... NOT COVERED
54: lit_int, replace `3` with `2`, at examples/simple/src/not_covered.rs@5:8-5:9(fn simple_assert_not_covered) ... NOT COVERED
55: binop_cmp, replace `<` with `<=`, at examples/simple/src/not_covered.rs@5:6-5:7(fn simple_assert_not_covered) ... NOT COVERED
56: binop_cmp, replace `<` with `>=`, at examples/simple/src/not_covered.rs@5:6-5:7(fn simple_assert_not_covered) ... NOT COVERED
57: binop_cmp, replace `<` with `>`, at examples/simple/src/not_covered.rs@5:6-5:7(fn simple_assert_not_covered) ... NOT COVERED
58: lit_int, replace `2` with `3`, at examples/simple/src/primetest.rs@5:11-5:12(fn primetest) ... killed
59: lit_int, replace `2` with `1`, at examples/simple/src/primetest.rs@5:11-5:12(fn primetest) ... killed
60: lit_int, replace `0` with `1`, at examples/simple/src/primetest.rs@5:16-5:17(fn primetest) ... killed
61: binop_eq, replace `==` with `!=`, at examples/simple/src/primetest.rs@5:13-5:15(fn primetest) ... killed
62: lit_int, replace `2` with `3`, at examples/simple/src/primetest.rs@6:20-6:21(fn primetest) ... killed
63: lit_int, replace `2` with `1`, at examples/simple/src/primetest.rs@6:20-6:21(fn primetest) ... killed
64: binop_eq, replace `==` with `!=`, at examples/simple/src/primetest.rs@6:17-6:19(fn primetest) ... killed
65: lit_int, replace `1` with `2`, at examples/simple/src/primetest.rs@8:12-8:13(fn primetest) ... killed
66: lit_int, replace `1` with `0`, at examples/simple/src/primetest.rs@8:12-8:13(fn primetest) ... killed
67: binop_eq, replace `==` with `!=`, at examples/simple/src/primetest.rs@8:9-8:11(fn primetest) ... killed
68: lit_bool, replace `false` with `true`, at examples/simple/src/primetest.rs@9:15-9:20(fn primetest) ... killed
69: lit_int, replace `3` with `4`, at examples/simple/src/primetest.rs@11:16-11:17(fn primetest) ... killed
70: lit_int, replace `3` with `2`, at examples/simple/src/primetest.rs@11:16-11:17(fn primetest) ... killed
71: binop_cmp, replace `<=` with `<`, at examples/simple/src/primetest.rs@12:16-12:18(fn primetest) ... killed
72: binop_cmp, replace `<=` with `>=`, at examples/simple/src/primetest.rs@12:16-12:18(fn primetest) ... killed
73: binop_cmp, replace `<=` with `>`, at examples/simple/src/primetest.rs@12:16-12:18(fn primetest) ... killed
74: lit_int, replace `0` with `1`, at examples/simple/src/primetest.rs@13:20-13:21(fn primetest) ... killed
75: binop_eq, replace `==` with `!=`, at examples/simple/src/primetest.rs@13:17-13:19(fn primetest) ... killed
76: lit_bool, replace `false` with `true`, at examples/simple/src/primetest.rs@14:19-14:24(fn primetest) ... killed
77: lit_int, replace `2` with `3`, at examples/simple/src/primetest.rs@16:13-16:14(fn primetest) ... killed
78: lit_int, replace `2` with `1`, at examples/simple/src/primetest.rs@16:13-16:14(fn primetest) ... SURVIVED
79: lit_bool, replace `true` with `false`, at examples/simple/src/primetest.rs@18:4-18:8(fn primetest) ... killed


65 mutants killed (6 by timeout)
14 mutants SURVIVED (7 NOT COVERED)
82.28% mutation coverage
```

</p>
</details>

In the long term, I hope to implement html-exports that in include all information in an accessible manner. However, I also want to be able to produce a meaningful report on stdout. Each line of the report is quite long, which is not optimal.

I would like to take feedback on the current format.

power-fungus:
I also implmented coverage analysis using the hints you gave in the review. If given the right parameters, the test suite will write coverage hits to a file, similar to the mutations. The runner enables coverage analysis by default and I currently dont see any reason make this optional.

llogiq:
Good progress! I'm fine with making coverage mandatory.

power-fungus:
I implemented the mutator `stmt_call`, which implements the removal of functioncalls and methodcalls that appear as statements ended with semicolons. Of course, I am aware of your blog posts about statement mutation and implemented the removal of statements minding the traps that you hinted at. 

The implementation of this mutator shows the extensibility of the proposed approach: Without much hassle, it was possible to add the mutation of other ast-constructs. Unfortunately, an issues appeared.

As you warned in the blog-post, the code `a(return b)` results in a compilation-error when removed. Specifically, this happens when all of the following conditions are met:
* a function is called with an argument of type `!`
* the function should return a non-unit value (a return type of `!` is a simmilar issue)
* there is no other exit-statement (end of block without `;` or `return` or `panic!`, or ...),  further down the line.

This is impossible to detect in the general case. Aside from the example above, following examples suffer from the same issue
  * `a(loop{})`
  * `a(panic!())`
  * `a(unimplemented!())`.
  * `a(x?);` with `x: Result<!,_>` once the `never_type` has been stabilized

I am unsure what to think of this problem. This only happens to code that is effectively dead and `rustc` prints a warning about that ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ea77a53de2577eb5678261be0b2df50a)). I expect that such code-patterns are very rare. In the next few weeks, I hope to apply `mutagen` to a set of larger libraries and check for compilation errors and we will see if this issue appears in practice. As far as I can tell, control-flow analysis is off-limits since it requires non-local information.

llogiq:
I think the compiler cannot make a distinction between non-local functions returning or not, but even if not taking `!` into account, changing control flow will open up a *huge* can of worms, because borrowck can and will ruin our day.

I suggest a second-best solution: remove just the outer call by wrapping it in a synthetized function that will or will not dispatch the call on mutation and call that instead.

This will only get rid of the call, not the statement, but it's better than nothing.

Control-flow and data-flow checking might still enable us to remove statements if we can assert that control flow is untouched and all data still initialized, though.

power-fungus:
When assuming that the function call is not dead code, then the mutation does not structurally change the control flow.

> I think the compiler cannot make a distinction between non-local functions returning or not, but even if not taking ! into account, changing control flow will open up a huge can of worms, because borrowck can and will ruin our day.

I am afraid, I do not underatand what you mean by that. 

A mutation that only removes the call but still computes the arguments is quite interesting.

llogiq:
What I mean is we need some analysis to see if removing the statement changes control flow and if removing it leaves uninitialized data that might be accessed later. The latter will likely lead to compile failures whereas it's uncertain where the former leads, but we want to avoid it anyway.

Removing a call while leaving the args intact is unusual but certainly possible, either by creating a wrapper `fn` and routing the call through that or by putting the args in local bindings first and `if`-ing the call.

power-fungus:
Addressing your second concern: Currently, we do not mutate assignment-statements like `x=1`. 

However, I am not sure if we need to avoid changing anything about the control-flow at all costs. In my opinion, if we are unable to find any examples where code breaks that do not involve dead code, we should use standard statement-deletion.

llogiq:
Perhaps we should start from a minimum viable mutation (that would be: remove argless call with unit return type) & work up from there (e.g. allowing `Expr::Path`s as args).

power-fungus:
I am trying to apply `mutagen` to existing libraries and found several issues. I track this progress in the repository [mutagen-applied](https://github.com/power-fungus/mutagen-applied)

Repostiories tested: chrono, rust-csv, actix, actix-web, indexmap. Some more to come

Issues found and fixed
* we are not allowed to mutate
	* const functions and declarations
	* patterns
	* const-expressions in array-lengths, i.e `[5u8; 4]` as expression
	* types, e.g. `[u8]`
* the mutation `*x == *y` fails to compile if x or y are not copy and needed elsewhere
* there is the folder `target/debug/deps`, from which test-suites are run, but are copied to `target/debug`
	* The runner looks for the exact same file in `./deps` relative to the test-executable. If found, this executable will be used instead

Issues found
* `edition2018` is required
	* for the existence of the path `::mutagen::...` without the use of `extern crate`
	* this is annoying since we not only require nightly and have some MSRV but also require an edition
* some integer-literals need an explicit type-suffix
	* the mutation of `fn f() -> u8 {1 + 2}` fails to compile
* The feature `#![feature(proc_macro_hygiene)]` may need to be enabled when letting `mutagen` run over code that defines a macro, i.e. `macro_rules` inside function

llogiq:
I think that requiring 2018 edition is acceptable, as updating isn't too hard.

The type suffix thing puzzles me – what does the mutated code look like?

power-fungus:
The root cause is the mutation of `+`. When only enabling `binop_add`, `fn f() -> u8 { 1 + 2 }` expands to:
```rust
fn f() -> u8 {
    ::mutagen::mutator::MutatorBinopAdd::run(
        1usize,
        1, 
        2,
        ::mutagen::MutagenRuntimeConfig::get_default(),
    )
}
```
And the compiler error is `type mismatch resolving '<i32 as std::ops::Add>::Output == u8'`


The root cause is: middle arguments of `run(_,1,2, _)`, which represent the original expressions are defaulted to `i32`, which causes a type-mismatch. [playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=f64336276e66eb7d2f763ea3a2583daa) to a simpler minimal example.

Somehow, the type-inference system is unable to determine the type of the literals. Most probably, this is defined behavior, although I am unsure. I am not aware of any way to tell the compiler to change this behavior.

llogiq:
I had the same problem with shifts (see my [blog post](https://llogiq.github.io/2018/04/11/shift.html) about it). I guess you should be able to reuse the solution.

power-fungus:
I did not see that blog post. I guess I have to dig deeper into the history. However, the issue described in the post only hints at one facet of the problem. 

Running example: `let x: u32 = (1 << 2) * 3` This compiles just fine. Replacing `<<` with its method, we get `1.shl(2) * 3`. Now, there are two issues:
* The compiler does not understand that `1.shl(2)` is supposed to be  `u32`
* When applying the trick described in the post, the compiler understands that `1.shl(2)` is indeed `u32`, but it still does not know that `1` should be `u32`. Minimal example: `let x: u32 = 1.shl(2)` or `fn f() -> u8 { 1.add(2) }` as described above.

As far as I know, this applies only to integer literals since everything else does not have defaulting rules (is this correct?). Literals have the unique property that all operations on them produce the same value (is this correct?).

In that case, I may have found a solution ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6aa759375a811a4e50c7f1e192e9bc0a)). `1 << 2` gets mutated to `{ let x = 1; if false { x } else {x.shl(2)}}`. Under the assumption that `<T as Shl>::Output == T`, the code should be correct. This version of the transformation can be enabled when a integer literals appears on the left side of an operation. 

To my delight, this also solves the issue of `(1 << 2) * 3` and similar cases I came up with. The reason might be that the guarantee that `1` and `1<<2` have the same type is the missing piece of information. This is also hinted at in the [comment above the section linked in your blogpost](https://github.com/rust-lang/rust/blob/a1c21ed7e21642b85947ba5d030bbaeffbe377de/src/librustc_typeck/check/op.rs#L103-L110).

Hopefully, the typechecker can be made stronger to detect these cases, although I am not sure that this is possible in a coherent manner.

llogiq:
Exactly, using `if false { /* original expression */ } else { /* mutated expression */ }` will give the typechecker the needed information.

The reason for this is a) the integer default to i32 and b) the special rules for operators.

power-fungus:
So, I've been done some further experiments and think reached a conclusion on the issue about type-inference for literals. In most cases, code can be generated such that the compiler accepts the mutated code, however, it might be the case that certain corner cases can never be mutated without compilation errors.

It seems that the trick using the if-expression can be replaced by using a different `run` function that sets all to the same type: `add_native_int<I: Add<I, Output=I>>(l: I, r: I) -> I`. This makes mutation a bit easier in some corner cases. (Playground)[https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=84a59bcfbe031f18501f2032a63fa003] for a small working example

The difficulty is to detect when to use the more general `add` function and when to use the specialized version `add_native_int`. From the perspective of the type-system, the answer is simple: if the left-side is known to be an integer-value, then the native-version should be used. Unfortunately, this is a non-local property in general. This (Playground)[https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b2943b804a65d6501df3cb7fadfa4a58] demonstrates this issue.

I try to implement a best-effort heuristic that detects when the left side of the add-expression is an integer but it must be clear that this is not detect all possible cases. The other arithmetic operations, bit operations, and negation all suffer from this issue and can be dealt with in the same manner.

I am willing to accept that some manual intervention is is required when applying mutagen to an existing codebase. Of course, we want to minimize the effort for manual intervention and it is not acceptable that common patterns of code cause problems. I feel that requiring to add integer-suffixes in certain situations is acceptable.

llogiq:
I think the `if` trick is better than trying to guess and accepting miscompilations, especially if the resulting errors are hard to pinpoint and fix.

However, if we absolutely cannot guarantee successful compilation, we should at least try to a) keep the error messages as understandable as possible and document ways to deal with them.

power-fungus:
The if-trick is not a general solution either. It requires that the left side of the `+` operation has the same type as the result of the addition, which is not the case in general. The example in the second [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=b2943b804a65d6501df3cb7fadfa4a58%5D) from my previous post, cannot be rewritten in terms of the if-trick since the add-impl from `Foo` outputs `u8`, in which case the original 

Furthermore, the current approach does not fix the `rhs` of the `add` operation to be the same value, although this can be fixed. Another advantage of the `run_native_int` approach over the if-trick is that it is not required to add extra-local variables, which introduces several new difficulties when correctly transforming the code. The requirement of the extra variable can be seen in the first [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=84a59bcfbe031f18501f2032a63fa003) from my previous post.

If there is a snippet that is handled well by the if-trick but the `add_native_int` approach fails, I will consider this solution instead.

power-fungus:
I implemented the heuristic about the `+` expressions. I would appreciate if you could do some experiments and see if you can come up with some unexpected results in the current setting. I will look how this heuristic performs on existing crates.

llogiq:
Will do, but it'll take a while. I need some uninterrupted time for such things, which I, alas, don't have until Wednesday.

llogiq:
Sorry for letting you wait so long. I think I found a solution: We need a third argument for our `add_sub` method that will have the type `<A as Add<Rhs=B>>::Output` (we could also use a free type `C` and possibly restrict it in a `where` clause). The result of the method will then also be `C` (or `<A as Add<Rhs=B>>::Output`, respectively).

power-fungus:
Your suggestion is very promising. However, the that approach duplicates the input expressions, which is not allowed when the expressions move something out of a variable. I've not been able to use this idea to get a working transformation in all cases.

power-fungus:
I am quite happy with the current layout of the report and consider this as done for now. For the time being, I am also happy with the mutators implemented so far. I implemented mutators for numeric operations `+`, `-`, `*` and `/` as well as the bit-operations `&`, `|`, `^`. 

After that, I would like to publish the current state as `mutagen-0.2`. I added you to the authors in the `Cargo.toml`s, which was not the case during my `mutagen-preview` proposal. Is this ok? If necessary, I would write a post about the release. I will also improve my repository `mutagen-applied` that tries to show how mutagen could be applied to existing libraries.

llogiq:
I'm currently very short on time, but will review soon. I'd also like to see some docs regarding open tasks for the future.

power-fungus:
I wrote a short list of in [beyond mutagen-0.2](https://github.com/power-fungus/mutagen/blob/master/docs/beyond-02.md). These tasks could give a roadmap for the mid-term future of mutagen.

llogiq:
:+1:

power-fungus:
During my work for [mutagen-applied](https://github.com/power-fungus/mutagen-applied), I found some more issues when applying mutagen to the crate `bytes`:

**There is a line `std::process::abort();` and its output type of `!` is required.**

The previos version rewrote the statement into `if [...] {std::process::abort();}`, which has type `()`.

I introduced a fix by generating the following code: `if [...] {std::process::abort();} else {::[...]::stmt_call_to_none()}`, which uses similar tricks to other optimistic mutators. 

This also fixes the issues discussed above for `f(return 1)`-like cases. Further, I labelled the mutator for deleting such statements as optimistic.

power-fungus:
I just realized a major issue. Currently, we suggest adding `#[cfg_attr(test, mutate)]` to the code under test. Then, we run all test suites found by `cargo test`. However, the `test` flag is enabled for the unit tests directly in the lib only. It is not enabled for other tests (e.g. files from `tests` folder).

As I understand it: **It is impossible for the macro `mutate` to be applied in integration tests while having `mutagen` in the `dev-dependencies` section**

I see several ways to resolve this. I am also happy to discuss other approaches.
* only consider unit-tests as executed by `cargo test --lib` for mutation coverage and stick with the current suggestions.
* suggest that `mutagen` is feature gated. This implies a few things: 
  * the crate has to introduce a new feature (e.g. `mutationtest`, suggested name can be discussed but cannot be named `mutagen`)
  * the dependency `mutagen` must be a real dependency, not just a `dev-dependency` and should be optional
  * the feature has to imply the dependency `mutagen` 
  * the `mutate` attribute has to be guarded with `#[cfg_attr(feature="mutationtest",mutate)]`
  * the runner must enable the feature for the tests, hence it must know the name of the feature either by convention or configuration
  * It is no longer guaranteed by cargo that no mutated code is shipped with production


short example for a possible suggested `Cargo.toml` configuration
```
[features]
mutationtest = ["mutagen"]
[dependencies]
mutagen = {optional = true}
```

I am not sure where to go with this and which direction is the least dangerous or most useful.

llogiq:
I would take the current approach and lobby the rust devs to set the `test` flag also when running integration tests.

In other news, David Tolnay has [documented a way to hack together specialization on stable](https://github.com/dtolnay/case-studies/tree/master/autoref-specialization). I'm currently trying to introduce this to overflower, and it would fit well with mutagen, too.

power-fungus:
I am fine with the current approach. I am currently refactoring the runner to respect coverage measures of each testsuite individually. 

However, I believe that the choice to disable the `test` flag and dev-dependencies inside the integration tests is based on the fact that they should test the crate "as any outside crate would". If possible, I will not do the lobbing necessary, but am happy with you or anyone else doing it.

I will try David Tolnay's method. However, we still require `proc_macro_span` feature for getting the filenames and locations for mutations. The tracking issue https://github.com/rust-lang/rust/issues/54725 has made little progress. If we need to be on nightly anyway, I would like to stick to the "official but unstable" approach instead. If it is possible to provide the desired functionality without any unstable features, I will invest the effort to cut them all.

power-fungus:
I uncovered another issue. (see Post in [users.rust-lang](https://users.rust-lang.org/t/unexpected-compilation-error-with-procedural-macros/34310)). This means that type inference rules can cause compilation errors in code that is not transformed by the mutagen. How should we deal with this?

power-fungus:
I did some improvements on the printed report and the running mechanism. My tests on [mutagen-applied](https://github.com/power-fungus/mutagen-applied) have been successful. Integrating mutagen into an existing project is not completely automatic and requires minimal amount of manual intervention in some cases. However, this effort is far smaller than the analysis of the mutationtest report, which is obviously a manual process.

*I would like to merge and publish my proposed version of mutagen as v0.2*. Is this possible within November?

llogiq:
I finally found the time to review this. Good job and thank you!

Issue #154, opened 2020-01-21
Feature Request: passing feature flags, and ignoring mutations


elichai:
Hi, 
First this crate is awesome :) Thank you!.
Second, In my view there are 2 important features missing to add this to a CI:

1. Feature gate the mutations, in real code you don't want to risk mutagen affecting the code with regular tests and the compilation overhead. 
(a naive solution is https://github.com/llogiq/mutagen/compare/master...TheBlueMatt:master a real one involves the ability of passing arbitrary feature flags when running `cargo mutagen`)

2. There are some mutations that can't be "fixed" with more tests, because their path isn't really reachable or the mutation doesn't really matter to the function, so a way to add "ignore" to mutations, either with a file (like sanitizers supressions) or with attributes (like `#[allow(...)]`) that can be also feature gated.

Thanks,
Elichai.

power-fungus:
Thank you for your feedback. 

Did I understand your first request correctly: "make it possible to run `cargo mutagen --features x,y,z` and `cargo mutagen --all-features`? I can do that.

The diff also includes a panic when a single mutation is not killed. Insisting on 100% mutation score is not the intention of mutation testing. However, I understand that regressions in mutation score can be caught in CI. Could you open a seperate issue for that?

The topic of skipping a mutations, should also be discussed in a separate issue. I have experimented with some design, but waited for feedback until publishing that featureset.

PS: what version of `mutagen` are you using?

elichai:
I agree about the panic, Altough it can make CI's easier if it fails on first mutation (assuming I can skip some)

And I'm currently using master.

power-fungus:
The first part is implemented in #155. Would it be possible to open separate issues for the remaining concerns?

PR #155, opened 2020-01-22
add options to pass features in mutagen runner

power-fungus:
Fixes first part of #154.

This adds flags `--features` and `--all-features` to the runner in the same way cargo handles them.

### DIFF
--- a/examples/feature-gated/Cargo.toml
+++ b/examples/feature-gated/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "feature-gated"
+version = "0.2.0"
+authors = ["Samuel Pilz <samuel.pilz@posteo.net>"]
+edition = "2018"
+
+[dev-dependencies]
+mutagen = {path = "../../mutagen"}
+
+[lib]
+doctest = false
+
+[features]
+with_mutagen = []
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,4 +13,7 @@ script:
 
   # run executable cargo-mutagen on test project
   - pushd examples/simple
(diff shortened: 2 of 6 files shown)


examples/feature-gated/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "feature-gated"
+version = "0.2.0"
+authors = ["Samuel Pilz <samuel.pilz@posteo.net>"]
+edition = "2018"
+
+[dev-dependencies]
+mutagen = {path = "../../mutagen"}
```
llogiq: This could be an optional dependency.

power-fungus: I tried that, but cargo responded with: "Dev-dependencies are not allowed to be optional: `mutagen`"

elichai: https://github.com/rust-lang/cargo/issues/1596

elichai: https://github.com/rust-lang/cargo/issues/1596

elichai:
Nice. Thank you!

examples/feature-gated/Cargo.toml
power-fungus: One thing to consider: Mutagen would then be included when running `cargo build --release --all-features`. However, in the current form such call would fail. 

Please note that this is an example and people will figure out how they want to use mutagen themselves. If anyone really feels strongly about this issue, I can adapt the example.

llogiq: Nah, in that case it's OK.

Interesting. I think that's a missing cargo feature.

PR #158, opened 2020-01-24
inference improvements

### LINKED ISSUE #157: Mutation literals corrupts types
Minimal reproducible example:
```
use mutagen::mutate;

#[mutate]
fn hi(mut ret: i64) -> i64 {
    let sh = if ret > 5 {
        0
    } else {
        8
    };
    ret += sh - 1;
    ret
}

fn main() {
    println!("Hello, world! {}", hi(500));
}
```

Without the macro it compiles fine. with the macro it returns the following:
```
error[E0277]: cannot add-assign `i32` to `i64`
  --> src/main.rs:11:9
   |
11 |     ret += sh - 1;
   |         ^^ no implementation for `i64 += i32`
   |
   = help: the trait `std::ops::AddAssign<i32>` is not implemented for `i64`
```

mutagen dep: git+https://github.com/llogiq/mutagen.git#99d9e12bf4e28e3a734b36f216650cec09a54c6c

power-fungus:
Fixes #157. The mutation of the snippet given in the issue used to compile but now fails to compile. This PR reverts to the previously use strategy, with slight alterations.

**Background**

We have 2 functions:

```rust
// does not work in old approach
#[mutate]
fn test1(mut ret: i64) -> i64 {
    1 + 2
}
// does not typecheck in new approach
#[mutate]
fn test1(mut ret: i64) -> i64 {
    let x = 1;
    x + 1
}
```

The first snippet did not typecheck correctly in the previous version. This led me to implementing a different approach based on detecting if the left side of an arithmetic operation is a integer literal or derived from some integer literal. This made the first one compile but was not smart enough to detect the second case. I thought I could extend it further, but it turned out that there is not enough information to solve this problem in all cases.

Previously, the trick was to rewrite `a + b` into `if false {a+b} else { <mutator-code>}`. 
I underestimated the potential of the `if false` trick, which is an error on my part. 

This PR switches to a modified version of the `if false` trick: `a+b` is transformed to `{let _tmp = a; if false {_tmp+b} else {<mutator-code based on _tmp and b>}}`. This makes both versions compile and also passed several other test cases.

**This PR is a work in progress. Feedback is welcome.**

### DIFF
--- a/mutagen-core/src/mutator/mutator_binop_bit.rs
+++ b/mutagen-core/src/mutator/mutator_binop_bit.rs
@@ -4,7 +4,7 @@ use std::convert::TryFrom;
 use std::ops::Deref;
 use std::ops::{BitAnd, BitOr, BitXor};
 
-use proc_macro2::{Span, TokenStream};
+use proc_macro2::Span;
 use quote::quote_spanned;
 use syn::spanned::Spanned;
 use syn::{BinOp, Expr};
@@ -70,25 +70,6 @@ pub fn run_xor<L: BitXor<R>, R>(
     }
 }
 
-pub fn run_native_num<I>(
-    mutator_id: usize,
-    left: I,
-    right: I,
-    original_op: BinopBit,
-    runtime: impl Deref<Target = MutagenRuntimeConfig>,
-) -> I
-where
-    I: BitAnd<I, Output = I> + BitOr<I, Output = I> + BitXor<I, Output = I>,
-{
-    runtime.covered(mutator_id);
-    let mutations = MutationBinopBit::possible_mutations(original_op);
-    if let Some(m) = runtime.get_mutation_for_mutator(mutator_id, &mutations) {
-        m.mutate(left, right)
-    } else {
-        original_op.calc(left, right)
-    }
-}
-
 pub fn transform(
     e: Expr,
     transform_info: &SharedTransformInfo,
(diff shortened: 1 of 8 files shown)


llogiq:
Why do you need the temporary? Does this work if the addition doesn't consume `a`? In that case moving `a` into a temporary could break code borrowing it later on.

power-fungus:
The temporary variable is necessary. Without it, the mutation of first snippet would not compile. 

Example ([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3896d4cf7da7e9ee4b98f561ed39d1fd))
We want to transform the expression `1+2`. The naive transformation replaces the`+` by some form of prefix `add` that contains the appropriate information to activate some mutation conditionally. The transformed code would look like `add(1,2)`. However, we know that this fails to typecheck since the first `1` is defaulted to `i32`. This is known and you proposed the `if false` trick to fix the problem.  The output of the `if false` for the statement `1+2` trick is `if false {1+2} else {add(1,2)}`. This does not fix the original problem: The type of the literal `1` is still defaulted to `i32`. The underlying issue is that the first and second `1` do not have the same type, which was one of the intentions of the `if false` trick. So I figured that a temporary variable will solve exactly that. 

I added some test cases and performed some smoke testing. It seems that the transformation is still valid in the face of non-copy types. Informally, this makes sense since `a` is consumed by the addition, by definition of the `Add` trait.

llogiq:
I'm on mobile right now, but you can implement `Add` on references.

llogiq:
It just occurred to me that we might create our own type to add. With rebalanced coherence we may be able to make the first example `1 + MutateAdd(..) + 2`.

power-fungus:
I do like the idea of inserting a custom type in the middle of the add operation. However, this requires to write `impl<T> Add<MutagenAdd> for T`, which is not allowed due to orphan rules.

Implementing `Add` on referencs is no issue here. If `impl Add for &Foo` then, `foo + foo` would not compile with or without mutation. `&foo + &foo` works in both cases.

I suggest merging this approach (which is a slight modification of your original idea).

power-fungus:
I also implemented the approach outlined above to mutators for bit-operations and negation

elichai:
> I do like the idea of inserting a custom type in the middle of the add operation. However, this requires to write `impl<T> Add<MutagenAdd> for T`, which is not allowed due to orphan rules.


Well you can manually impl it for all primitives: https://play.rust-lang.org/?gist=132755b74337dc5ca1dd0c7e9179c26b

power-fungus:
> Well you can manually impl it for all primitives: https://play.rust-lang.org/?gist=132755b74337dc5ca1dd0c7e9179c26b

I did not know that. However, the transformation has to be valid for all type combinations, e.g. `string + "foo"` and for custom types.

The more general issue is that the compiler has a special case for type inference in the case of integer expressions. Any expression like `MutagenAdd + 1 + 2` or `MutagenAdd(1) + 2` will fall short in type inference. The `if false` approach tricks the compiler to do the special casing for us.

power-fungus:
I would consider this PR ready for merging. I have added some more testcases that show that the improvements fix a all cases I could think of.

Hints to any code that still breaks in this new system is welcome.

power-fungus:
Relevant parts of this PR were merged together with #160

PR #172, opened 2020-10-19
Pass through --package and --workspace arguments

ogham:
I have a project with multiple crates in a workspace, with only the "inner" one needing mutation testing. The data in `target/mutagen` was generated when I ran `cargo test --workspace`, but cargo-mutagen didn't detect it unless I also passed `--workspace` to the `cargo test` that cargo-mutagen runs, so I made this patch to add the argument.

---

Commits:
* This commit adds the `--package` and `--workspace` arguments to cargo-mutagen. They're implemented in the same way as the `--features` and `--all-features` arguments, getting passed through to the `cargo test` invocation.

### DIFF
--- a/mutagen-runner/src/main.rs
+++ b/mutagen-runner/src/main.rs
@@ -32,6 +32,14 @@ struct Options {
     /// Activate all available features
     #[structopt(long)]
     all_features: bool,
+
+    /// Package to run tests for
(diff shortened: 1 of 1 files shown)


llogiq:
this looks good to me, I just have one question, but that shouldn't block merging. I'd like to see some documentation on how to use this, but that may well be a follow-up.

PR #180, opened 2021-09-16
Fix clippy and update dependencies

barafael:
Thanks for making this proof of concept library! It's already useful. I noticed there are some clippy warnings, fixed them, and updated the dependencies, FWIW. Let me know what you think and if there is something else I can do here to get this merged.


llogiq:
Thank you!

Issue #182, opened 2022-05-06
Add Github Action


royrustdev:
# Add Github Action for Continuous Integration

## Expected Behavior

Implementation of Github Action would be very helpful to maintain the project

## Current Behavior

Github Action is missing.

## Solution

I am going to implement this. Please take a look when I submit.

royrustdev:
Github action is added. This issue #182 is now closed.

PR #183, opened 2022-05-06
#182 Add Github Action

royrustdev:
I implemented Github Action on the project.

Please let me know if something needs to be added

Thanks


llogiq:
Thank you. I'll see if I need to do anything to activate the workflow.

PR #189, opened 2022-07-23
Add str literal mutation

samgoldman:
If you're open to new mutations, I created a mutator for string literals. It replaces empty strings with a single character string and for non-empty strings it does a few things: append and prepend a single character as well as replacing the string with an empty string. I'd welcome any feedback. Thanks!

### DIFF
--- a/docs/mutators.md
+++ b/docs/mutators.md
@@ -33,6 +33,23 @@ Byte-literals like `b'a'` are not mutated by this mutator.
 
 Customization is WIP
 
+## lit_str
+
+### Target Code
+
+`&str` literals.
+
+Char literals like `'a'` are not mutated by this mutator.
+
+### Mutations
+
+* If not empty:
+  * Replace the literal with an empty string
+  * Prepend `'-'`
+  * Append `'-'`
+* If empty:
+  * Replace the literal with `"A"`
+
 ## unop_not
 
 ### Target Code
--- a/docs/mutators.md
+++ b/docs/mutators.md
@@ -45,7 +45,8 @@ Char literals like `'a'` are not mutated by this mutator.
 
 * If not empty:
   * Replace the literal with an empty string
-  * Replace the literal with `"-"` (or if the literal is already exactly `"-"`, with `"*"`)
+  * Prepend `'-'`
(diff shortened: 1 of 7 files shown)


samuelpilz:
This seems like a really good implementation of a new mutation, thank you. Documentation and tests are well done.
The only thing that concerns me is the owned String as return value from the `run` function. Generally, changing the type of an expression is tricky, it messes with type-inference and could result in different trait resolutions.

One potential issue in particular: Does this implementation break code like this (because the owned String is only a temp variable)?

```rust
fn a() -> usize {let x = ""; x.len()}
```

I suggest using a return value of `&'static str`. This requires baking the possible mutation literals into the source code.

samgoldman:
Thanks for the feedback, that makes sense. I made the update, although I had to remove the appending and prepending mutations. I also added a test case for the example you provided and it doesn't break it (and it didn't break before changing to `&'static str` either).

samuelpilz:
Indeed, for some reason `{let x = &String::new(); x.len()}` works (for some reason, the lifetime of the tempvar gets extended). I thought this was a minimal example of lifetime extension. Probably a better example is:

```rust
fn a() -> &'static str {""}
```

This code should not break as well (which is now fixed).


I really liked the string-dependent mutations. My idea is changing the signature of `run` roughly to:

```rust
fn run(
    mutator_id: usize,
    original_lit: &'static str,
    mutations: &[&'statc str],
    runtime: &impl Deref<Target = MutagenRuntimeConfig>,
)  -> &'static str {...}
```
This allows you to create the mutated strings at compile-time and have them available at run-time as `&'static str`

samgoldman:
Yeah, the new example would have broken the original version. I updated to the signature you suggested and that example works (and I've added it as a test case).

samuelpilz:
Great. One last comment: could you revert the documentation to your original proposal?
Otherwise, this is ready for merge (although I don't have the rights)

samgoldman:
I forgot to leave a comment but that should be done.

llogiq:
Thank you! I didn't have the time to look at this until now, but Samuel has already given all the feedback needed.

Thanks to both of you!

